Issue #53, opened 2020-03-12
C7 Variable reference point 2 still needed?


kerimoyle:
From here: https://cellml-specification.readthedocs.io/en/cellml-2-drafting/reference/specC5_variable_reference.html

> C7. Variable reference
> 1. When present in an information item which is a descendant of a component element, a variable reference SHALL be the name of a variable, and SHALL refer to the variable element in that component with a name attribute identical to the variable reference.
> 2. In all other cases, a variable reference SHALL consist of a component reference and a variable name. In this case, the variable reference SHALL be treated as if it was just the variable name present in the component element referenced by the component reference.

Just wondering about 2 - where are these other cases?

MichaelClerx:
:thinking:

nickerso:
Connections. 1 is essentially dealing with variables in math and 2 is dealing with variables in connections. (I'm sure I'm forgetting others but those are the main ones springing to mind right now).

kerimoyle:
Kiiiiiinda ... but it's not clear that the "component reference" is the `component_1` or `component_2` in the `connection` element?  I was expecting something like `<something variable="myComponent:myVariable">` from the wording.  If it's only in connections can we just say that?

MichaelClerx:
@kerimoyle is this one fixed now?

kerimoyle:
Yes :)

PR #71, opened 2020-03-17
Updated variable reference section i53

kerimoyle:
For discussion around #53 ... I've had a first go at rewording section C5. I'm not sure it's any clearer though! Please emend as needed.

### DIFF
--- a/reference/sectionC_interpretation.rst
+++ b/reference/sectionC_interpretation.rst
@@ -335,16 +335,22 @@ Component reference
 Variable reference
 ------------------
 
-#. When present in an information item which is a descendant of a
-   :code:`component` element, a variable reference SHALL be the name of a
-   variable, and SHALL refer to the :code:`variable` element in that
-   component with a :code:`name` attribute identical to the variable
-   reference.
+#. When present in a :code:`math` element a variable reference SHALL be the
+   name of a variable, and SHALL refer to the :code:`variable` element in the
+   parent :code:`component` with a :code:`name` attribute identical to the
+   variable reference.
 
-#. In all other cases, a variable reference SHALL consist of a component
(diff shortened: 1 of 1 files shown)


nickerso:
I think worth keeping in mind that this source was generated from the google doc via a complicated process :) I agree worth thinking about formatting going forward to make it easy to see what is changing and should make an issue for that.

agarny:
> Thanks for taking the first go at this @kerimoyle. I personally find the revised version more confusing :) but it is also incorrect (e.g., a `math` element will never have a variable reference). I'll see if I can come up with an alternate revision.

See issue #72.

kerimoyle:
> (e.g., a `math` element will never have a variable reference). 

This is what I was talking about yesterday ... What *is* a variable reference then?  I (clearly) don't even know ... I thought it included the variable name between the <ci> tags? otherwise how is it needed in the maths blocks (see conversation [here](https://github.com/cellml/cellml-specification/issues/53#issuecomment-598590552)? )

kerimoyle:
> since we are "starting" work on this repo

@agarny ... while you might be "starting" ... others have been working for a month and hope to be "finishing" pretty soon! :)

agarny:
> > since we are "starting" work on this repo
> 
> @agarny ... while you might be "starting" ... others have been working for a month and hope to be "finishing" pretty soon! :)

One month is nothing considering that CellML has been around for 20 years. That was my point. :)

kerimoyle:
> Thanks for taking the first go at this @kerimoyle. I personally find the revised version more confusing :) but it is also incorrect (e.g., a math element will never have a variable reference). I'll see if I can come up with an alternate revision.

Ping @nickerso  ... thoughts on revision?

Issue #72, opened 2020-03-17
Source code formatting


agarny:
The idea is to ensure that our reST code is easily maintainable. This means (to be completed, if needed):
 - [x] Have one sentence per line (this makes it much easier to do a `diff`).

kerimoyle:
Unfortunately, the rst linter complains about lines which are more than 80 characters long.  This is why changing one word somewhere can have a knockon effect on the rest of the paragraph ... :(

MichaelClerx:
Can we not disable that somehow?

MichaelClerx:
Uhm, locally I can build documentation with longer lines just fine. Is there some extra tool involved beyond sphinx?

kerimoyle:
It's not a thing with building, it just gives me warnings locally.  If anyone is volunteering to go through and change everything into one line per sentence (it wasn't like that coming from the google doc!) they are more than welcome ...

MichaelClerx:
Sure!

kerimoyle:
Hey @MichaelClerx  ... so far so good, but do you know why the magical conversion stopped partway through some of the files?  Have a look at `reference/sectionB_elements.rst` for an example?

MichaelClerx:
Hmmm, will have another go!

MichaelClerx:
Not sure what happened there, either I missed a massive chunk or something happened with merging / some new PR overwrote things?

kerimoyle:
Not sure either. I originally thought I'd screwed it up with a PR, but couldn't find where it had happened if so ... (doesn't mean it wasn't me somewhere else though!).  Thanks for fixing :)

kerimoyle:
Closing as now in practice, fixed since #80

Issue #77, opened 2020-03-18
PDF version of pure formal spec only


kerimoyle:
For the purists, a PDF version of the normative spec only (ie: without the notes from the informative sections) needs to be available.

kerimoyle:
Apparently this isn't possible through readthedocs (though is possible elsewhere) to automagically serve two separate PDF documents, so we'd need to have separate branches (boo) to enable people to be able to download a "manual-type" document of the annotated spec, if we wanted that.  What do others think?  
Option 1:  PDF download is formal spec only, no notes.
Option 2: PDF download of formal spec AND PDF download of annotated/informative spec too? This will mean either a) maintaining two branches only for this purpose, or b) serving a static copy of the formal spec as a PDF, and allowing download of the informative spec as the locally-built PDF?

agarny:
I would probably go with option 1. Easier to maintain. Also, people who want/need the notes will probably be just fine with just an HTML-based version (I probably would be).

MichaelClerx:
Is readthedocs going to be the official way we spread it anyway? Or is that a development thing?
It could end up a bit weird with versioning and such, and possibly lead to lots of slightly different versions of the spec floating around?

kerimoyle:
I'm struggling to get the PDF and HTML versions to match.  The Latex writer insists on numbering the chapters, thus introducing an extra place in the overall numbering of headings.  I've tried various combinations and fixes as recommended in the sphinx/rtd docs, but alas, not getting very far.  Does anyone have any experience/thoughts/advice/whatever ?  Here's what I'm talking about:

HTML version:  Uses "Section A" and then numbers
![image](https://user-images.githubusercontent.com/46887220/77264480-4945e880-6cff-11ea-9818-f16ae2561739.png)

PDF version: Uses "Chapter 1: Section A" and then numbers
![image](https://user-images.githubusercontent.com/46887220/77264556-75616980-6cff-11ea-9f59-0c1ca4f5f710.png)

Thus, A.2.1 has become 1.2.1.  If we're going to stick with this higher-level thing (the sections A, B, etc) then that effectively adds another level of division, so we have (down the tree) 5 levels of points/sub-points if these chapter numbers stay ... which isn't great.  Any ideas?

> Is readthedocs going to be the official way we spread it anyway? Or is that a development thing?
> It could end up a bit weird with versioning and such, and possibly lead to lots of slightly different versions of the spec floating around?

[...]

PR #80, opened 2020-03-19
Changed more text to one-sentence-per-line format.

MichaelClerx:
More #72


nickerso:
Seems to be an issue with building the PDF which gives a "build failed" error on readthedocs, but the HTML does seem to be updating. So I think safe to merge this in...

kerimoyle:
Yeah, that's expected, and comes from my work-in-progress stuff to get it to build two different PDF versions ... I had to get the big file rearrangement in before disappearing over the weekend though.  See https://github.com/cellml/cellml-specification/issues/77#issuecomment-602264239

PR #68, opened 2020-03-17
Addresses i60, i63 re imports of units and components

kerimoyle:
This is my attempt at rewording sections C01 imports, C02 unit references and C04 component references in order to clarify the issues in #60 and #63.  Please feel free to emend as you see fit!

PR #86, opened 2020-03-23
I60 i63 imports

kerimoyle:
Replaces my messed up #68


kerimoyle:
Sorry @MichaelClerx ... I screwed up the previous PR #68 :(  This is the same, but with the changes you suggested.  I'll leave the other open so we can see the difference...

reference/sectionC_interpretation.inc
```diff
@@ -29,7 +35,7 @@ Units reference
 
 #. The units referencing rules are:
 
-   #. Where there is a :code:`units` element with a :code:`name` attribute identical to the units reference, then the units reference SHALL refer to that :code:`units` element.
+   #. Where, within the same infoset, a :code:`units` element has a :code:`name` attribute identical to the units reference, the units reference SHALL refer to that :code:`units` element.
```
MichaelClerx: So perhaps to replace the old "no rule held to be followed" bit we should just add a line at the end saying "no other values of the name attribute are valid"?

MichaelClerx: Although I see we don't have anything like that for the component refs.

I guess it would be ok to leave this out if we go by the rule that the spec says everything that is valid (stuff it doesn't say is invalid per definition)

reference/sectionC_interpretation.inc
```diff
@@ -12,8 +12,14 @@ Section C: Interpretation
 Interpretation of imports
 -------------------------
 
-#. Each :code:`import` element present in a :ref:`CellML infoset<specA_cellml_infoset>` (the importing infoset) SHALL define a new and separate instance of the CellML infoset referenced by the :code:`href` attribute (the imported infoset).
-   See :ref:`Units reference<specC_units_reference>` and    :ref:`Component reference<specC_component_reference>` for the specifics    of importing units and components.
+#. Each :code:`import` element present in a :ref:`CellML infoset<specA_cellml_infoset>` (the importing infoset) SHALL define a new and distinct instance of the CellML infoset which is referenced by the :code:`href` attribute (the imported infoset).
+   This imported infoset instance SHALL provide the context in which the child of an import element (a :code:`units` or :code:`component` element) is defined and interpreted.
+
+#. The importing of an element SHALL be considered recursively such that items referenced or imported by the :code:`import` element child (a :code:`units` or :code:`component` element) are themselves available to that :code:`import` element child.
```
MichaelClerx: "referenced or imported" --> "referenced"

I think it just matters that everything referenced by the imported item is available to that item (no matter how it got in the imported items scope)

reference/sectionC_interpretation.inc
```diff
@@ -236,7 +242,7 @@ Component reference
 
 #. A component reference present in an information item which is a descendant of a :code:`model` element SHALL be identical to either the :code:`name` attribute on a :code:`component` element or to the :code:`name` attribute on an :code:`import component` element.
 
-#. A component reference which is identical to the :code:`name` attribute on a :code:`component` element SHALL be treated as a reference to that :code:`component` element.
+#. Where, within the same infoset, a :code:`component` element has a :code:`name` attribute identical to the component reference, the component reference SHALL refer to that :code:`component` element.
 
 #. A component reference which is identical to the :code:`name` attribute on an :code:`import component` element SHALL be treated for the purposes of referencing as if the component reference appeared in the imported model, and referred to element with the :code:`name` specified in the :code:`component_ref` attribute of the :code:`import component` element.
```
MichaelClerx: Further simplification:

"but only to the explicitly specified :code:`import` element child, unless that child is itself explicitly imported."
can be
"but only to explicitly specified :code:`import` element children."
?

kerimoyle:
Rendered version here: https://cellml-specification-dev.readthedocs.io/en/i60_i63_imports/

reference/sectionC_interpretation.inc
kerimoyle: I'll delete it.

reference/sectionC_interpretation.inc
kerimoyle: Done.

reference/sectionC_interpretation.inc
kerimoyle: ps - Need to check around this area (C.1.2 - 3) as it might be duplicated by a merge ...?

reference/sectionC_interpretation.inc
```diff
@@ -12,8 +12,14 @@ Section C: Interpretation
...
+#. Each :code:`import` element present in a :ref:`CellML infoset<specA_cellml_infoset>` (the importing infoset) SHALL define a new and distinct instance of the CellML infoset which is referenced by the :code:`href` attribute (the imported infoset).
+   This imported infoset instance SHALL provide the context in which the child of an import element (a :code:`units` or :code:`component` element) is defined and interpreted.
+
+#. The import of an element SHALL be considered recursively such that items referenced by the :code:`import` element child (a :code:`units` or :code:`component` element) are themselves available to that :code:`import` element child.
+   It is noted, for the avoidance of doubt, that the importing infoset SHALL NOT have access to such recursively imported items, but only to explicitly specified :code:`import` element children.
+
+#. The import of an element SHALL be considered recursively such that items referenced by the :code:`import` element's child (a :code:`units` or :code:`component` element) are themselves included.
```
kerimoyle: Not sure if this is still needed??

MichaelClerx: How so?

kerimoyle: The two points seem really similar? not sure if the second is a leftover previous version of the first?

MichaelClerx: Oh yeah. I agree drop 2nd one

kerimoyle: Done in 6078884

reference/sectionC_interpretation.inc
kerimoyle: OK, done.

reference/sectionC_interpretation.inc
```diff
@@ -8,8 +8,13 @@
...
 -------------------------------------
 
-#. Each :code:`import` element present in a :ref:`CellML infoset<specA_cellml_infoset>` (the importing infoset) SHALL define a new and separate instance of the CellML infoset referenced by the :code:`href` attribute (the imported infoset).
-   See :ref:`Units reference<specC_units_reference>` and :ref:`Component reference<specC_component_reference>` for the specifics of importing units and components.
+#. Each :code:`import` element present in a :ref:`CellML infoset<specA_cellml_infoset>` (the importing infoset) SHALL define a new and distinct instance of the CellML infoset which is referenced by the :code:`href` attribute (the imported infoset).
+   This imported infoset instance SHALL provide the context in which the child of an import element (a :code:`units` or :code:`component` element) is defined and interpreted.
+
+#. The imported infoset SHALL be considered holistically such that items relevant to the imported infoset (:code:`unit` children of imported :code:`units` elements or encapsulation children of imported :code:`component` elements) are themselves available to that :code:`import` element child.
```
kerimoyle: comprehensively?

kerimoyle: #. The imported infoset SHALL be considered holistically such that items relevant to the imported infoset (:code:`unit` children of imported :code:`units` elements or encapsulat**ed :code:`component`** children of imported :code:`component` elements) are themselves available to that :code:`import` element child.

Issue #64, opened 2020-03-16
_ref vs reference?


kerimoyle:
We throw around the terms "units reference" and `units_ref`, and "component reference" and `component_ref` ... but they don't seem to be consistent (or at least, I'm confused ...)  

**Q0:** Can we add a rider to the beginning of C2 and C4 just saying what we're talking about?

"A "component reference" refers to the `component_ref` attributes used in `encapsulation` and `import component` items, and the `component_1` and `component_2` attributes used in `connection` items."

"A "units reference" refers to the `units_ref` attribute in `import units` items, and the `units` attribute used in `unit` and `variable` items."

> C4: Component reference
> A component reference SHALL be the **name of a component**, and SHALL be interpreted based on the **context within the CellML model in which it occurs**.

**Q1:** `component_ref` attributes are used in `encapsulations` and `imports`;`component_1` and `component_2` attributes are used in `connections`... but are all of these covered under the "component reference" title? In which case, I think we need to separate the imports from the others, as otherwise the "context within the CellML model in which it occurs" is a bit weird?

> C2: Units reference
> A units reference SHALL be **a CellML identifier** and SHALL be interpreted dependent on the context of the CellML model in which it occurs, according to the units referencing rules defined later in this section.

**Q2:** Why is this a CellML identifier when the component reference has "a component name".  Should they be the same, and if so, which is better?

> 1. A units reference SHALL be a CellML identifier and SHALL be interpreted dependent on the context of the CellML model in which it occurs, according to the units referencing rules defined later in this section.
> 2. A CellML infoset MUST NOT contain a units reference for which no referencing rule can be held to have been followed.
> 3. The units referencing rules are:
>    1. Where there is a units element **??where??** with a name attribute identical to the units reference, then the units reference SHALL refer to that units element.
>    2. Where there is an import units element in the CellML infoset, such that the import units element has a name attribute identical to the units reference, then the units reference SHALL be treated with respect to referencing rules as if the units reference appeared in the imported infoset, and referring **(referred?)** to the name specified in the units_ref attribute of the import units element.
>    3. Where the units reference is equal to the value in the ‘Name’ column of the Built-in units table, then the units reference SHALL be a reference to the built-in units corresponding to that row of the table.

**Q3:** Can we rearrange the order of these points to make it clearer without referring to "later in this section" stuff? eg: move 1 to be 3.1, 2 to after 3?

nickerso:
[...]

Issue #45, opened 2020-03-09
Radians and sterradians


MichaelClerx:
Currently defined as meter/meter?

https://libcellml-tutorials.readthedocs.io/en/pr344_documentation/reference/specC2_units_reference.html

- First, isn't that an incomplete unit reduction? I'd expect this to be simplified to {}
- Secondly, if we're going to be really strict about them, shouldn't these both be base units?

jonc125:
It's because the radian is defined in SI as the ratio of two lengths, and CellML follows SI for the built-in units (mostly).

MichaelClerx:
But that still wouldn't give us that as a unit reduction, right?

MichaelClerx:
> Where the set of tuples consists of tuples which have the same name element, those tuples SHALL be combined into a single tuple with that name element and an exponent being the sum of the combined tuples’ exponents. If the resulting tuple’s exponent term is zero, the tuple SHALL be removed from the set of tuples. Note that this can result in the set of tuples being empty.

So there's no way to distinguish between the ratio of two lengths and plain old dimensionless in CellML

nickerso:
If you used libCellML to compare radians and dimensionless, then they would be the same. In the past people have expressed the opinion that it is useful to keep these ratio type units as having their actual definition as per SI. Strain is often the other one that is a length ratio that some people use as dimensionless and some use as m/m.

MichaelClerx:
That's good! But we still can't have "metre/metre" as the units reduction of radians, because it's not a units reduction :D

jonc125:
It can still be the definition though, just not in fully reduced form :)

MichaelClerx:
So what we're saying is that, _in CellML_, radian == sterradian == dimensionless.
But then, uniquely, we use the spec to say that they're a kind of dimensionless with an interesting backstory?

MichaelClerx:
For the other derived SI variables we don't give a backstory, e.g. `pascal = N / m^2`. We just give the reduction, as that's the only meaningful thing in CellML, and the only useful thing for reading the spec.

I mean, I don't _want_ radians and sterradians to be new base units as that'd be another feature I need to implement in Myokit and cellmlmanip. But it makes no sense for me to hide this CellML deficiency by saying "radian == dimensionless but it used to not be". Let's just own up to it :D

kerimoyle:
[...]

PR #87, opened 2020-03-23
issues 64 and 45: built-in units table and units references

kerimoyle:
See #45 and #64


kerimoyle:
Rendered version here: https://cellml-specification-dev.readthedocs.io/en/i64_ref_vs_reference/reference/formal_only/specC02_units_reference.html#table-built-in-units

kerimoyle:
Haha, I thought about it ... and I knew whichever way I did it wouldn't be the right one!

kerimoyle:
Updated now, see: https://cellml-specification-dev.readthedocs.io/en/i64_ref_vs_reference/reference/formal_only/specC02_units_reference.html#table-built-in-units

agarny:
> Updated now, see: https://cellml-specification-dev.readthedocs.io/en/i64_ref_vs_reference/reference/formal_only/specC02_units_reference.html#table-built-in-units

Much, much better in my view! A few more things though:
 - Why do the symbol-based colums have contents in italics? It makes them stand out, which I am not sure we want...? Ok, I have just realised that it's because you use the `:math:` construct. Do we want to do that? 
 - I would probably center-align the symbol-based columns. At least, to see how it looks.
 - I would use "⋅" rather than "." in things like "N⋅m" (note the order of "N" and "m", i.e. "N⋅m" which seems to be more common rather than "m⋅N").
 - Should we really put different variants for the common equivalent definitions? E.g., wouldn't "N⋅m" be sufficient for Joule?

kerimoyle:
> Why do the symbol-based colums have contents in italics? It makes them stand out, which I am not sure we want...? Ok, I have just realised that it's because you use the :math: construct. Do we want to do that?

I tried it both ways and I think this is better.  If not, we have to deal with ohms, and it just looks weird having some maths in an equation, and some not ... so I changed them all over.  I think that they're being rendered as italic and bold though, which isn't so great ... 

> I would probably center-align the symbol-based columns. At least, to see how it looks.

I'll put that on the bottom of the priorities heap for now ... I'll get content correct first! 

> I would use "⋅" rather than "." in things like "N⋅m" (note the order of "N" and "m", i.e. "N⋅m" which seems to be more common rather than "m⋅N").

Good idea.

> Should we really put different variants for the common equivalent definitions? E.g., wouldn't "N⋅m" be sufficient for Joule?

I've just followed the examples in the SI units derived units definition from [here](https://en.wikipedia.org/wiki/SI_derived_unit).

agarny:
Fair enough / agreed regarding points 1, 2 and 4.

reference/sectionC_interpretation.inc
```diff
@@ -62,7 +62,7 @@ Table: Built-in units
 | farad         | :math:`F`         | (metre, -2), (kilogram, -1),     | | :math:`C/V`                      |
 |               |                   | (second, 4), (ampere, 2)         | | :math:`s/{\Omega}`               |
 +---------------+-------------------+----------------------------------+------------------------------------+
-| gram          | :math:`g`         | 0.001 * (kilogram, 1)            |                                    |
+| gram          | :math:`g`         | :math:`0.001\cdot` (kilogram, 1) |                                    |
```
agarny: Hmm... the rendering is going to look a bit odd. Part of it is going to be in italics. Why not simply have `0.001⋅(kilogram, 1)`?

kerimoyle: How do you get the dot without :math: ?  Please feel free to change as you'd like to - these tables take *ages* to make tiny adjustments to!

agarny: Just write verbatim what I wrote above.

kerimoyle: Where is the dot symbol?? I can copy yours but I want to know how you got it in the first place?

kerimoyle: OK, done.  I still want to know how to get that dot though!

agarny: From [here](https://en.wikipedia.org/wiki/International_System_of_Units#Derived_units). It corresponds to the [dot operator](https://en.wikipedia.org/wiki/Multiplication_sign#Unicode_and_HTML_entities).

kerimoyle:
Review please @agarny @nickerso ?

reference/sectionC_interpretation.inc
```diff
@@ -23,6 +23,8 @@ Interpretation of imports
 Units reference
 ---------------
 
+A "units reference" refers to the :code:`units_ref` attribute in :code:`import units` items, and the :code:`units` attribute used in :code:`unit` and :code:`variable` items.
```
agarny: - Why the `"` around "units reference"? I can see that it's also done in C4, but should we really do it? (Just asking as I think I might be fine with it.)
- Wouldn't it be more accurate to have " ... refers to **the value of** the :code:`units_ref` attribute... and **the value of** the :code:`units` attribute..."? (If so, we ought to update C4 too.)
- Technically speaking, we don't `import units` items, just `import` items for units. (Same for C4.)

reference/sectionC_interpretation.inc
```diff
@@ -42,83 +44,87 @@ Units reference
...
+| **Name**      | **Symbol**        | | **Unit reduction tuple**                | **Common definitions**             |
+|               |                   | | multiplier⋅(base, exponent)             |                                    |
++---------------+-------------------+-------------------------------------------+------------------------------------+
+| ampere        | :math:`A`         | \-                                        |                                    |
++---------------+-------------------+-------------------------------------------+------------------------------------+
+| becquerel     | :math:`Bq`        | (second, -1)                              | :math:`1/s`                        |
++---------------+-------------------+-------------------------------------------+------------------------------------+
+| candela       | :math:`cd`        | \-                                        |                                    |
++---------------+-------------------+-------------------------------------------+------------------------------------+
+| coulomb       | :math:`C`         | (second, 1), (ampere,1)                   | | :math:`s⋅A`                      |
```
agarny: I can see that, in commit 333b6f7, you replaced `\cdot` with `⋅` in `math` constructs. I don't think you should do that. It messes up the rendering. `⋅` should only be used in "normal" text, not within a `math` construct.

reference/sectionC_interpretation.inc
```diff
@@ -232,6 +238,8 @@ yocto    −24
 Component reference
 -------------------
 
+A "component reference" refers to the :code:`component_ref` attribute used in :code:`import component` items, to the :code:`component_ref` elements which occurs as children of :code:`encapsulation` items, and to the :code:`component_1` and :code:`component_2` attributes used in :code:`connection` items.
```
agarny: See remarks in https://github.com/cellml/cellml-specification/pull/87/files#r398161216.

reference/sectionC_interpretation.inc
kerimoyle: 1. Because I got really confused my first time reading this section.  I didn't know what the phrase "units reference" meant - this is supposed to define it, hence the ".  If there's a clearer way am happy to change it?
2,3. You're right, I think ... will update.

reference/sectionC_interpretation.inc
kerimoyle: Done in cff1dc9

reference/sectionC_interpretation.inc
kerimoyle: Done in cff1dc9

reference/sectionC_interpretation.inc
kerimoyle: I haven't seen the difference?  I was hoping that your magic dot would render smaller and with less space than the \cdot symbol, but as they're both coming out the same I gave up!  I've searched for a smaller latex symbol than \cdot, but can't seem to find anything that looks good ... if you have a suggestion am happy to fix it, but because these tables are *such* a massive timesink, I really don't want to just play around ...

reference/sectionC_interpretation.inc
```diff
@@ -238,7 +238,7 @@ yocto    −24
 Component reference
 -------------------
 
-A "component reference" refers to the :code:`component_ref` attribute used in :code:`import component` items, to the :code:`component_ref` elements which occurs as children of :code:`encapsulation` items, and to the :code:`component_1` and :code:`component_2` attributes used in :code:`connection` items.
+The term "component reference" refers to the value of a :code:`component_ref` attribute used in an :code:`import` item for :code:`component` items, to :code:`component_ref` elements which occur as children of :code:`encapsulation` items, and to the value of :code:`component_1` and :code:`component_2` attributes used in :code:`connection` items.
```
agarny: Why is the wording different for units and component? (E.g., "the value of **the**" vs. "the value of **a**")

reference/sectionC_interpretation.inc
agarny: All I am saying is that we should use the `:math:`  construct the way it is intended to be used, i.e. using `\cdot` (or something else of the sort although I am not sure there is something better for what we want to do here) and let the renderer do its thing, and `⋅` for "normal" text because that's the best we can do in that case.

reference/sectionC_interpretation.inc
kerimoyle: I think it should be indefinite, so will go with **a**.

kerimoyle: Done in 1f6e180

reference/sectionC_interpretation.inc
kerimoyle: Right, I've nuked it all.  Much cleaner now, I hope (see 9fa5100) If not, please make whatever changes you need to ... this table has shortened my life enough!

kerimoyle:
Ah, bother, I can't figure out how to make magic dot render properly in the PDF :|  I'm going to add that to the list of things to sort out in #95 (re PDF rendering) rather than here, as I think it's better there and this one should really go in.  So, (finally) re-re-re-re (etc) view please @agarny @MichaelClerx

reference/sectionC_interpretation.inc
agarny: > this table has shortened my life enough!

:)

agarny:
> Ah, bother, I can't figure out how to make magic dot render properly in the PDF :| I'm going to add that to the list of things to sort out in #95 (re PDF rendering) rather than here, as I think it's better there and this one should really go in.

That seems reasonable to me.

Issue #56, opened 2020-03-13
Initial value as variable - do we still need it?


MichaelClerx:
CellML 1.1 introduced having variables as `initial_value` attributes:

> The value of the initial_value attribute may be a real number or a variable. A modeller may wish to store the initial and boundary conditions in a separate file that is imported by the current model.

So now you could do e.g.
```
<model>
  <comp>
    <var x initial_value="y"
    <var y public_interface="in">
    <math>
      dx/dt = ...
```
and then have an imported CellML file that sets `y`, along with all the other initial values

In CellML 2.0 you can do this more easily, as
```
<model>
  <comp>

    <var x interface="public">
    <math>
      dx/dt = ...
```
because with the loss of directionality you can set an initial value anywhere in the connected variable set.

So... do we still need the initial-value-is-a-variable construct? It'd reduce the burden on implementers, and probably simplify resets too

agarny:
Good question! I would definitely be in favour of getting rid of that feature since it would indeed make things easier, e.g. with regards to code generation (I was actually going to check whether we support it and do it, if needed). @nickerso?

MichaelClerx:
@jonc125 do you have thoughts on this?

nickerso:
We don't *need* the initial value as a variable in 2.0, but this is a significant change and will impact model translation where people are following "best practice" and using this feature in CellML 1.1 models to separate the math from the parameters. Coming up with a translation which respects this separation may need more than an XSLT...

MichaelClerx:
Are many people using the feature? I've never once had an problem due to not supporting it

agarny:
> We don't _need_ the initial value as a variable in 2.0, but this is a significant change and will impact model translation where people are following "best practice" and using this feature in CellML 1.1 models to separate the math from the parameters. Coming up with a translation which respects this separation may need more than an XSLT...

So, your concern is about the conversion from CellML 1.1 to CellML 2.0. If so, then I wouldn't consider this as being a valid reason for not dropping the initial value as a variable. As you confirmed, that feature is not needed in CellML 2.0. So, it should be dropped. Simple as that. (That's why we are going from 1.x to 2.x., to allow "big" changes.)

nickerso:
[...]

Issue #58, opened 2020-03-13
Initial value cycles


MichaelClerx:
I think we're missing a rule that says initial_values can't create cycles?

    <var A init="B">
    <var B init="A">

MichaelClerx:
Or is that just "silly maths" again?

agarny:
This is valid CellML, just "silly maths" indeed.

MichaelClerx:
And then there's `<var A init="A">`

(all these examples are in valid PseudML, just in case you're wondering)

agarny:
PseudML?

MichaelClerx:
Oh dear. That one bothers me actually.
You can write
```
<math>
  x = x
</math>
```
because that just says that x equals itself, which is fine.

But the initial value rule says

> Where the initial_value attribute is a variable reference, it SHALL be interpreted as a statement that the variable on which the attribute appears is equal to the referenced variable under the conditions when the initial value holds.

that is something very different from what we do with the maths!

So

1. Stuff in mathml: Equations which hold eternally. Might be rubbish if your model is silly.
2. Stuff in initial_value attributes: Statements about variable equivalence (!!! that needs rewording, at the very least) at the "initial point"

MichaelClerx:
It _might_ be that changing the wording to "value of" in two places fixes this. Not sure.

So then "maths = eternal truths", "initials = truths about the initial condition" ? And both can be nonsense without affecting validity?

nickerso:
> So then "maths = eternal truths", "initials = truths about the initial condition" ? 

Yes. I think so.

> And both can be nonsense without affecting validity?

exactly.

kerimoyle:
@MichaelClerx do you have wording in mind for this one?  I'll add a comment to the informative spec too.

MichaelClerx:
I think it's entirely for the informative spec this one!
I'd just write that nothing in CellML prohibits you from writing "a has initial value b" and "b has initial value a" or similar constructs. That this is valid CellML just like x = 1/0 but just not a useful thing to do

kerimoyle:
Updated in #88 - see https://cellml-specification-dev.readthedocs.io/en/i56_initial_variables/reference/formal_and_informative/specC06_interpretation_of_initial_values.html for rendered version.

PR #88, opened 2020-03-24
I56 initial value as a variable - update informative section

kerimoyle:
Addresses #56 

I've updated the "see more" section of the C.6 page with an example model.  Thoughts?  I'm not convinced what I've said is correct, but it's a place to start from. ...   The rendered version is here:
https://cellml-specification-dev.readthedocs.io/en/i56_initial_variables/reference/formal_and_informative/specC06_interpretation_of_initial_values.html


reference/informative/informC06_interpretation_of_initial_values.rst
```diff
@@ -8,4 +8,117 @@
 
   .. container:: infospec
 
-    **TODO**
\ No newline at end of file
+    Best practice suggests that the :code:`initial_value` attribute is used to set the initial values only, rather than to simply set the value of a :code:`variable`.  
+    Initialising a variable is different from setting the value of a variable; the former is true only at the *beginning* of a simulation, whereas the latter is true *throughout* the simulation.
+    The option to use a variable reference for initialisation is provided in CellML 2.0 to enable easier migration from CellML 1.1 models, and may not be supported in future releases.
```
MichaelClerx: Thanks! This is looking good but there's quite a lot going on here!
Maybe we could separate it into a few key points, in different paragraphs (possibly with headings?)

- semantics intial value true at t0 vs equation true always
- init used to set constants is not recommended but possible
- init can now be on connected var, allowing init vals to be set in a separate component (possibly imported)
  - init = var can be used to set init vals in separate component (possibly imported) but not recommended

kerimoyle: Thanks :) I don't quite understand how to put this bit into practice:

> init can now be on connected var, allowing init vals to be set in a separate component (possibly imported)
> init = var can be used to set init vals in separate component (possibly imported) but not recommended

So ... init = val in external component, no init through the rest of the connected set? So my "good" example is not really so good ... how should I change it?

kerimoyle:
Included info on #58 here too, as it's in the same block.  Hopefully example is correct and wording clearer, but please suggest if not ... :)

reference/informative/informC06_interpretation_of_initial_values.rst
MichaelClerx: I'd just do an ODE with a component
```
component clever_stuff
    variable x
    variable y
    math:
        dx/dt = ...
        dy/dy = ...
component initialisation
    variable x initial_value=3
    variable y initial_value=7
```
?

reference/informative/informC06_interpretation_of_initial_values.rst
```diff
@@ -8,4 +8,148 @@
...
+
+    - A :code:`variable` whose value needs to be set for only the *beginning* of a simulation is initialised using the :code:`initial_value` attribute. 
+      
+        - This is most frequently used for state variables (those whose value is found by solving a differential equation).
+        - It's possible - but not recommended - to use a variable reference with the :code:`initial_value` attribute.
+          This option remains in CellML 2.0 only to provide ease of migration from CellML 1.1 models, and may be discontinued in future versions.
+          The recommended best practice is to make use of connections between variables to move all "hardcoding" of parameters (including initial values and constants) into a top-level component for ease of access.
+          This is demonstrated in the example below. 
+      
+    - A :code:`variable` whose value is *constant* throughout the simulation is best set within the :code:`math` block, rather than via the :code:`initial_value` attribute.  This is so that its value will be held to be true *throughout* the simulation, rather than only the beginning.
```
MichaelClerx: New line!

reference/informative/informC06_interpretation_of_initial_values.rst
```diff
@@ -8,4 +8,148 @@
...
+
+        Best practice for constants, variables, and initial conditions
+    
+
+    - A :code:`variable` whose value needs to be set for only the *beginning* of a simulation is initialised using the :code:`initial_value` attribute. 
+      
+        - This is most frequently used for state variables (those whose value is found by solving a differential equation).
+        - It's possible - but not recommended - to use a variable reference with the :code:`initial_value` attribute.
+          This option remains in CellML 2.0 only to provide ease of migration from CellML 1.1 models, and may be discontinued in future versions.
+          The recommended best practice is to make use of connections between variables to move all "hardcoding" of parameters (including initial values and constants) into a top-level component for ease of access.
```
MichaelClerx: Maybe we can just leave this sentence out? I don't think it's necessarily best practice to do this, just that _if you want to have them in a separate component_, then this is the best way to do it

reference/informative/informC06_interpretation_of_initial_values.rst
kerimoyle: Yep, got it, thanks.

reference/informative/informC06_interpretation_of_initial_values.rst
kerimoyle: OK, will do.

kerimoyle: Done in 6a718c4

reference/informative/informC06_interpretation_of_initial_values.rst
kerimoyle: Done in 6a718c4

reference/sectionB_elements.inc
```diff
@@ -359,7 +359,7 @@ A :code:`math` element information item (referred to in this specification as a
 
 .. container:: issue-math-cn-units-attribute
 
-   4. Any MathML :code:`cn` elements MUST each have an attribute in the :ref:`CellML namespace<specA_cellml_namespace>`, with a local name equal to :code:`units`. The value of this attribute MUST be a valid units reference.
+   4. Any MathML :code:`cn` elements MUST each have an attribute in the :ref:`CellML namespace<specA_cellml_namespace>`, with a local name equal to :code:`units`. The value of this attribute MUST be a valid :ref:`units reference<specC_units_reference>`.
```
agarny: Two sentences, so two lines.

kerimoyle: Done in 770e756

Issue #89, opened 2020-03-24
Change contact address and attribution


kerimoyle:
Add help@physiomeproject.org to the contacts

kerimoyle:
Closed by #90

nickerso:
Just to reopen this one, is there a reason not to use `editors@cellml.org` as the contact address? @agarny @hsorby @kerimoyle @MichaelClerx  (sorry not to think about this earlier...)

agarny:
Hmm... good point! Doh! Definitely +1 for `editors@cellml.org`.

kerimoyle:
Closed by #98

PR #90, opened 2020-03-24
I89 contact email

kerimoyle:
Addresses #89


reference/sectionA_definitions.inc
```diff
@@ -4,8 +4,8 @@
 
 **Authors:**
 
-Michael T. Cooling\*
 Michael Clerx
```
MichaelClerx: I like this of course, but should be not have a chat about first/last authors at some stage?

kerimoyle: I just changed the order to be alphabetic as he was no longer the corresponding author ... discussion welcome though :)

agarny: I agree with an alphabetical order, but using the first name. Thanks. 😉

More seriously, an alphabetical order using the last name is what I would expect here.

PR #95, opened 2020-03-25
i77 Generate PDF of formal spec

kerimoyle:
Addresses #77 

This PR generates a single PDF which represents only the formal text of the normative specification, and does not include any of the informative spec blocks, where the text and heading numbers in the PDF are identical to those in the HTML.

- [x] Figure out where to put the preamble stuff ...
- [x] Get the interpunct character to render in the Built-in units table PDF version, see #87 
- [ ] Get preamble address correct?  Do we need a DOI? #104
- [ ] Find a Latex wizard to help with formatting once the content (all other PRs) are finalised.

Rendered version is here: https://cellml-specification-dev.readthedocs.io/en/i77_pdf_of_formal_spec_only/index.html


MichaelClerx:
I'm trying to try this out but running into trouble getting this branch on my machine locally.

I've added your repo to my git, but then what...?

kerimoyle:
If you've forked my repo (kerimoyle/cellml-specification), you should be able to pull and push from your own fork as normal?  If you've already forked _this_ repo (cellml/cellml-specification) then you should be able to see this PR as a branch you can refer to? Perhaps referred to as something like "pr/95" or by its name?

nickerso:
> but then what...?

We work very similar to libCellML. Take a look at 
https://www.libcellml.org/en/latest/dev_setup.html#forking-your-own-copy - you'd need to add a remote to @kerimoyle's repo and fetch the branch from there...I don't think the pull request exists as a branch in this repo (cellml/cellml-specification).

kerimoyle:
No changes to content, only getting the PDF to render nice(r)!  I'd like to get this one in so that I can start on addressing #110 without too many conflicts ... There will be more to come on the PDF front, but I'll do that after the headings have changed so we can better see what it looks like.  So ... ok to merge? @agarny @nickerso @hsorby ?  Here's the rendered PDF at present:
 https://cellml-specification-dev.readthedocs.io/_/downloads/en/i77_pdf_of_formal_spec_only/pdf/

reference/informative_preamble.rst
```diff
@@ -0,0 +1,43 @@
...
+and Maxwell L. Neal.
+
+The authors also wish to acknowledge the significant contribution of the (discontinued) draft CellML 1.2 specification, much of the text of which was incorporated into this specification, although the semantics have changed considerably.
+The CellML 1.2 specification was itself the result of a collaborative effort by a number of researchers during 2008-2011\:
+
+Andrew K. Miller (who wrote the text reused here),
+Randall Britten,
+Jonathan Cooper,
+Alan Garny,
+Peter J. Hunter
```
hsorby: missing comma

formal.rst
```diff
@@ -1,18 +1,53 @@
...
 
+**Authors:**
+Michael Clerx,
+Michael T. Cooling,
+Jonathan Cooper,
+Alan Garny,
+Keri Moyle,
+David P. Nickerson,
+Poul Nielsen,
+and Hugh Sorby.
```
agarny: Shouldn't the "and" be on the previous line? (Same below.)

index.rst
```diff
@@ -1,18 +1,16 @@
 .. _index:
 
 ==========================================
-Normative specification for CellML 2.0
+CellML 2.0 Specification
 ==========================================
```
agarny: I would remove some "=" to make it the same length as the header.

agarny: Same below.

nickerso: or leave it...I was planning to put the normative spec bit back in once things settle down ;-)

reference/informative_preamble.rst
```diff
@@ -0,0 +1,43 @@
+.. _inform_preamble:
```
agarny: No way to have some kind of include file? It looks like it overlaps quite a bit (entirely) with `formal.rst`.

kerimoyle:
@nickerso It is there twice, because the PDF renderer doesn't recognise the interpunt symbol, so they're a little different between the `..only:: html` and `..only:: latex` versions.

reference/informative_preamble.rst
kerimoyle: True, but there are differences in the order of where things will appear, and eventually they will be more distinct as we get more information about the informative spec in the inform_preamble file.  I've found that using includes does odd stuff to the numbering and structure of the PDF, so need to keep that side as is for now.

formal.rst
kerimoyle: It's all rendered to a single line anyway ...

index.rst
kerimoyle: Will do elsewhere throughout later on

formal.rst
kerimoyle: 034c2ef done

reference/informative_preamble.rst
kerimoyle: 034c2ef done

agarny:
> @nickerso It is there twice, because the PDF renderer doesn't recognise the interpunt symbol, so they're a little different between the `..only:: html` and `..only:: latex` versions.
> 
> Ah, dang, you're right ... it's there twice in the PDF ... oops

This aside, can't we find a way around the interpunt symbol? I really don't like the idea of having that table twice (not good when it comes to maintenance).

kerimoyle:
@agarny Please do!  I've spent a day trying and failing ... :\  (hence most of the massive commit history of this branch!!).

kerimoyle:
It's also now defaulting to a `.` in both tables, which is something new ... previously it was simply omitted.  If that behaviour is stable then we can just remove the second table ...

kerimoyle:
OK, double table is fixed, new pages added in the PDF before both prefix and built-in units tables.  Only outstanding thing here are general formatting (length of title decoration) which will be done in another PR, and the interpunt thing (which will be done when pigs fly).  I'm going to merge this now and will add the rest as needed. ...

MichaelClerx:
Awesome!

PR #97, opened 2020-03-26
I53 variable reference rewording2

kerimoyle:
This PR replaces #71 whose base is too out of date.  Addresses #53.


kerimoyle:
@nickerso I've had another go, this time following the language/format of the other "reference" sections for units and components as in #87 .  Please reword as needed (am pretty sure it's still wrong as I'm confused about the maths etc ...)

PR #99, opened 2020-03-27
Tweaking author and contributor lists

nickerso:
I think this accurately recognizes people's contributions to the CellML 2.0 specification.


conf.py
```diff
@@ -1,37 +1,45 @@
 # -*- coding: utf-8 -*-
 #
-# CellML Specifications documentation build configuration file, created by
-# sphinx-quickstart on Tue Nov 13 11:59:15 2012.
+# libCellML documentation build configuration file, created by
```
agarny: I guess this is the result of some copying/pasting... I mean, it used to read `CellML Specifications` not `libCellML`. (I haven't checked the rest of this PR, but who knows what else got inserted by mistake.)

nickerso: sorry, this was a PR against the wrong branch. But perhaps that is an actual issue in the current codebase?

agarny: Don't know, haven't checked. Saw it because it's at the beginning of the PR, that's all. :)

Issue #105, opened 2020-03-29
Format headings with monospace font in section 3?


kerimoyle:
Currently we mix monospace font and normal throughout the headings:
![image](https://user-images.githubusercontent.com/46887220/77859234-a5ab8980-7264-11ea-8756-06d350e13674.png)
Should the headings in Chapter 3 be adjusted to match the others, and perhaps changed to avoid plurals of monospace ... eg: 
- Interpretation of `import` elements
- `units` references
etc ... ?

agarny:
Yes, I would do that.

kerimoyle:
Closed by #108

PR #108, opened 2020-03-29
I105 monospace headings

kerimoyle:
Making the heading fonts consistent in Chapter 3 as in Chapter 2.  Addresses #105


reference/sectionC_interpretation.inc
```diff
@@ -315,8 +315,8 @@ Interpretation of mathematics
 
 .. _specC_interpretation_of_encapsulation:
 
-Interpretation of encapsulation
--------------------------------
+Interpretation of ``encapsulation``
+-----------------------------------
```
hsorby: I think this should be
```
Interpretation of ``encapsulation`` elements
--------------------------------------------
```

reference/sectionC_interpretation.inc
```diff
@@ -398,8 +398,8 @@ Interpretation of map_variables
 
 .. _specC_interpretation_of_variable_resets:
 
-Interpretation of variable resets
----------------------------------
+Interpretation of ``variable`` ``reset`` elements
```
hsorby: I think this should be
```
Interpretation of ``reset`` elements
```

kerimoyle: I agree ... but it used to be "Interpretation of resets" ... then we added the "variable" quite recently? Not sure why though ...

kerimoyle: Done in 3bf4167

reference/sectionC_interpretation.inc
kerimoyle: Done in 3bf4167

reference/sectionC_interpretation.inc
```diff
@@ -339,8 +339,8 @@ Interpretation of encapsulation
 
 .. _specC_interpretation_of_map_variables:
 
-Interpretation of map_variables
--------------------------------
+Interpretation of ``map_variables``
```
hsorby: This is an element too right?

kerimoyle: Yeah, I guess so.  In the beginning I was only adding "elements" to the titles to avoid having plurals in the monospace font (eg:  `variable`s or `variables`).  I guess I should just make them all long and elementy though.

agarny: When editing the title, it would be good to add the corresponding number of `-`. I don't believe it's strictly needed, but it looks "better" when looking at the source code.

kerimoyle: Should the "Interpretation of mathematics" section now become "Interpretation of `math` elements"?

agarny: I would say so.

kerimoyle: Both done in 906fe31

agarny: Oops, my bad, I should have checked the commit!

Issue #123, opened 2020-03-31
Clean up references section


kerimoyle:
Need to make the inclusion of the references chapter consistent with the others so that latex doesn't complain about nested `toctrees`.

kerimoyle:
Closed by #127

Issue #36, opened 2020-03-09
Unicode reference


kerimoyle:
We currently include references to unicode symbols via the fileformat.info site (eg: [here](http://www.fileformat.info/info/unicode/char/0041/index.htm)).  This was written in 1993 ... I'm wondering whether we need to define the version we're using (through perhaps the unicode site itself?), and/or update to something this millennium?

kerimoyle:
Ping @MichaelClerx @nickerso @agarny @hsorby

hsorby:
I can't get to the reference for the symbol anywhere on the unicode site whilst the fileformat.info takes me directly to the definition.  I find the fileformat.info location the most helpful but as you say quite out of date.  I was trying to find similar info for version 5 as that is the ISO standard but cannot.

hsorby:
I think that letter hasn't changed since version 1 1993 so it has stayed the same further versions then build upon older versions?

kerimoyle:
Yeah, I'm guessing that none of the characters we care about are likely to change all that much ... though am now wondering whether they're better specified as UTF-8 if that's really [the default for XML](http://www.utf-8.com/)?

kerimoyle:
OK, a bit more digging shows that they the fileformat.info site doesn't update when newer versions are released, it just adds the new characters (eg: if we ever needed a [llama](https://www.fileformat.info/info/unicode/char/1f999/index.htm), we're in luck, it was added in 2018).  So I guess we're fine, I'll close this now.  Thanks @hsorby!

hsorby:
Oh we have to change the spec. we definitely need a llama!

kerimoyle:
My thought too ... sigh ... perhaps it will find its way into a tutorial or two ;)

MichaelClerx:
You can use them in URLs too http://michaelclerx.com/%f0%9f%a5%83/

PR #127, opened 2020-03-31
I123 references cleanup

kerimoyle:
Addresses #123. General reorganisation of the references section, added link to unicode.

Rendered version is here: https://cellml-specification-dev.readthedocs.io/en/i123_references_cleanup/reference/index_sectionD.html


reference/sectionD_references.inc
```diff
@@ -0,0 +1,17 @@
+
+.. _sectionD:
+
+Extensible Markup Language (XML) 1.1 (Second Edition) http://www.w3.org/TR/2006/REC-xml11-20060816 (16 August 2006)
+
+Mathematical Markup Language (MathML) Version 2.0 (Second Edition) http://www.w3.org/TR/2003/REC-MathML2-20031021/ (21 October 2003)
```
agarny: Prefer https://www.w3.org/TR/2003/REC-MathML2-20031021/.

reference/sectionD_references.inc
```diff
@@ -0,0 +1,17 @@
+
+.. _sectionD:
+
+Extensible Markup Language (XML) 1.1 (Second Edition) http://www.w3.org/TR/2006/REC-xml11-20060816 (16 August 2006)
```
agarny: Prefer https://www.w3.org/TR/2006/REC-xml11-20060816/. Otherwise 29 September 2006.

reference/sectionD_references.inc
```diff
@@ -0,0 +1,17 @@
...
+
+Mathematical Markup Language (MathML) Version 2.0 (Second Edition) http://www.w3.org/TR/2003/REC-MathML2-20031021/ (21 October 2003)
+
+Namespaces in XML 1.1 (Second Edition) http://www.w3.org/TR/2006/REC-xml-names11-20060816/ (16 August 2006)
```
agarny: Prefer https://www.w3.org/TR/2006/REC-xml-names11-20060816/.

reference/sectionD_references.inc
```diff
@@ -0,0 +1,17 @@
...
+
+RFC 2119: Key words for use in RFCs to Indicate Requirement Levels http://www.ietf.org/rfc/rfc2119.txt (March 1997)
```
agarny: Prefer https://www.ietf.org/rfc/rfc2119.txt.

reference/sectionD_references.inc
```diff
@@ -0,0 +1,17 @@
...
+
+Unicode 1.1.0  https://www.unicode.org/charts/ (June 1993)
+
+XML Information Set (Second Edition) http://www.w3.org/TR/2004/REC-xml-infoset-20040204/ (4 February 2004)
```
agarny: Prefer https://www.w3.org/TR/2004/REC-xml-infoset-20040204/.

reference/sectionD_references.inc
```diff
@@ -0,0 +1,17 @@
...
+
+XML Linking Language (XLink) Version 1.0 http://www.w3.org/TR/2001/REC-xlink-20010627/ (27 June 2001)
```
agarny: Prefer https://www.w3.org/TR/2001/REC-xlink-20010627/.

reference/sectionD_references.inc
```diff
@@ -0,0 +1,17 @@
...
+
+Extensible Markup Language (XML) 1.1 (Second Edition) http://www.w3.org/TR/2006/REC-xml11-20060816 (16 August 2006)
+
+Mathematical Markup Language (MathML) Version 2.0 (Second Edition) http://www.w3.org/TR/2003/REC-MathML2-20031021/ (21 October 2003)
+
+Namespaces in XML 1.1 (Second Edition) http://www.w3.org/TR/2006/REC-xml-names11-20060816/ (16 August 2006)
+
+RFC 2119: Key words for use in RFCs to Indicate Requirement Levels http://www.ietf.org/rfc/rfc2119.txt (March 1997)
+
+Unicode 1.1.0  https://www.unicode.org/charts/ (June 1993)
```
agarny: If we are to use that URL, then we should refer to Unicode 13.0 and the date should be 6 February 2020.

kerimoyle: see discussion on initial issue at #36 ?

reference/sectionD_references.inc
kerimoyle: See url - date looks like August 16?

reference/sectionD_references.inc
agarny: If we were pointing to https://www.fileformat.info/info/unicode/version/1.1/index.htm then, sure, Unicode 1.1.0 and June 1993 would make sense, but here we are pointing https://www.unicode.org/charts/, which explicitly mentions Unicode 13.0 and 6 February 2020... So, we can use that URL with the version and date of another.

reference/sectionD_references.inc
agarny: Sure, but edited in place on 29 September 2006. @nickerso and @hsorby, views?

reference/sectionD_references.inc
kerimoyle: 00b9efa

reference/sectionD_references.inc
nickerso: why is this not https://www.w3.org/TR/MathML2/ ? Surely if an updated MathML 2 spec was released we would be using it...

nickerso: and a reference list would generally have the date the reference was accessed for an online reference, not the date the resource was published.

reference/sectionD_references.inc
nickerso: as for MathML (and all these URLs), should be using the canonical reference not a versioned instance: https://www.w3.org/TR/xml11/. And if you want to include a date it would be the date accessed. But I'd suggest not putting a date.

reference/sectionD_references.inc
nickerso: https://www.w3.org/TR/xml-names11/

reference/sectionD_references.inc
agarny: > why is this not https://www.w3.org/TR/MathML2/ ? Surely if an updated MathML 2 spec was released we would be using it...

I was merely suggesting replacing `http` with `https`, not thinking about the URL itself, but I agree that https://www.w3.org/TR/MathML2/ is better.

> and a reference list would generally have the date the reference was accessed for an online reference, not the date the resource was published.

Not quite sure what you mean by this. Are you saying that we should have something like 1 April 2020 instead?

reference/sectionD_references.inc
agarny: Ok for not putting a date. After all, the URL is really all we need (and agreed re using the canonical reference).

reference/sectionD_references.inc
agarny: Indeed.

reference/sectionD_references.inc
agarny: Prefer https://www.w3.org/TR/xml-infoset/ instead.

reference/sectionD_references.inc
agarny: What should we do about this one? https://www.w3.org/TR/xlink/ refers to XLink 1.1...

reference/sectionD_references.inc
nickerso: > Not quite sure what you mean by this. Are you saying that we should have something like 1 April 2020 instead?

Yep, you'd normally have (Accessed: 1 April 2020) or similar. But I don't think we need dates as discussed in the line 4 thread :)

reference/sectionD_references.inc
nickerso: hmmmm....looks like we could switch to XLink 1.1 and everything should be fine...still uses the same xlink namespace.

Or we just stick to referencing the XLink 1.0 spec via that version specific URL and play it safe :)

kerimoyle:
Sorry, I'm getting lost in all the comments ...  I had a quick look for formatting options, and both the access date and the date of writing were included in a lot of them.  I suspect that for our purposes the date of writing is more valuable?  (ps - I'm very happy for someone who knows what's going on to take over this PR ... or give me some clear instructions, I can't figure out what's really wanted?)

agarny:
> Sorry, I'm getting lost in all the comments ... I had a quick look for formatting options, and both the access date and the date of writing were included in a lot of them. I suspect that for our purposes the date of writing is more valuable? (ps - I'm very happy for someone who knows what's going on to take over this PR ... or give me some clear instructions, I can't figure out what's really wanted?)

Personally, I would edit reference/sectionD_references.inc to read:

```
.. _sectionD:

Extensible Markup Language (XML) 1.1 https://www.w3.org/TR/xml11

Mathematical Markup Language (MathML) Version 2.0 https://www.w3.org/TR/MathML2/

Namespaces in XML 1.1 https://www.w3.org/TR/xml-names11/

RFC 2119: Key words for use in RFCs to Indicate Requirement Levels https://www.ietf.org/rfc/rfc2119.txt

Unicode 13.0.0 https://www.unicode.org/versions/Unicode13.0.0/

XML Information Set https://www.w3.org/TR/xml-infoset/

XML Linking Language (XLink) Version 1.1 https://www.w3.org/TR/xlink11/
```

Note that I have removed the edition information (when there was one) in case a canonical link was ever to reference another edition. Also, note that the link for Unicode 13.0(.0) is now completely different. Unicode 13.0.0 was officially released on 10 March 2020.

Otherwise, I can see that the original references page used to have "inline" links, e.g.

```
`Extensible Markup Language (XML) 1.1 <https://www.w3.org/TR/xml11>`__
```

Why is that not the case anymore?

agarny:
If people agree with https://github.com/cellml/cellml-specification/pull/127#issuecomment-607474889 then, sure, I am happy to implement the changes, but then I won't officially be allowed to review this PR anymore. :)

kerimoyle:
> Otherwise, I can see that the original references page used to have "inline" links, e.g.

`Extensible Markup Language (XML) 1.1 <https://www.w3.org/TR/xml11>`__
Why is that not the case anymore?

The point of rewriting the references section was to be able to provide links to people who might want to print it.  If you have inline links then the url doesn't get displayed.

kerimoyle:
There you go :)

agarny:
> The point of rewriting the references section was to be able to provide links to people who might want to print it. If you have inline links then the url doesn't get displayed.

I imagine that it relates to the PDF version? If so then I thought that we had "agreed" that inline links (to the spec) were fine? So, surely that includes URLs?

kerimoyle:
The only place this page is seen is when someone explicitly clicks on the Refernces tab, or prints the PDF.  Personally I'd rather see the address spelled out in both versions - with redirects etc it can be difficult to know where you're meant to be going unless you know the original address.  And it's a reference section: the url _is_ the reference.

agarny:
Ok, fine by me.

Issue #129, opened 2020-03-31
Should we define comments in the spec?


kerimoyle:
See: https://cellml-specification-dev.readthedocs.io/en/update_to_cellml2/reference/formal_and_informative/specA2.html#semantically-equivalent-cellml-infosets

We talk about "comment information items" in 1.2.3.1.1 but don't define them anywhere?

agarny:
I imagine that we are talking about an XML comment? If so, maybe this is the terminolgy we should be using, as well as link to https://www.w3.org/TR/2006/REC-xml11-20060816/#sec-comments?

kerimoyle:
Yes, an XML comment.  Then a follow-on, should we also be defining the characters (or character combo) used to create comments (!, <, > etc) too in 1.1 Terminology?

agarny:
Hmm, we probably ought to, I guess.

Otherwise, in 1.1 Terminology, the CellML 2.0 namespace is `http://www.cellml.org/cellml/2.0#`, but its link is to `http://www.cellml.org/cellml/2`...?

kerimoyle:
Yeah, I know, it's fixed in my latest.

MichaelClerx:
I'm pretty sure I brought up something about this in the google doc spec! There are 11 types of information item defined here https://www.w3.org/TR/xml-infoset/#infoitem (which we link to), and we designate 8 of them "specific". We're not really defining them just saying _of the 11 that exist_, we're going to now say something about these 8. And then we talk about the other 3 in veiled terms a bit, so I suggested at one point to make it more explicit...

MichaelClerx:
I think my point was that the stuff about changing CellML without affecting was about the other 3 types.

Main takeaway here is that we don't _define_ any of the 11 types in the CellML spec, we refer to already defined things from the XML spec, so we don't need to define comment information items, but could link to https://www.w3.org/TR/xml-infoset/#infoitem again the first time they're mentioned maybe

hsorby:
> but could link to https://www.w3.org/TR/xml-infoset/#infoitem again the first time they're mentioned maybe

I agree with this no need to define it, just reference it.

nickerso:
agreed.

Issue #118, opened 2020-03-30
Informative specification update


kerimoyle:
Since the informative sections have had no revision yet, we will need to tidy them up (and complete in places!).

- Check for hardcoded references and update
- Check that examples match the updated normative spec
- Grammar/formatting/punctuation etc.

- [x] Chapter 1
- [x] Chapter 2
- [x] Chapter 3

Also: defining comments?  #129

PR #132, opened 2020-03-31
I118 informative cleanup

kerimoyle:
Addresses #118 to update the informative sections of the html spec.

Also will close (when I know how):
- #129


kerimoyle:
(facepalm) wrong branch ... again ...

Issue #138, opened 2020-04-01
Informative 3.4


kerimoyle:
Write the "see more" block for this section.

PR #151, opened 2020-04-02
i138 Informative 3.4 References to component elements

kerimoyle:
Addresses #138 Informative block for the 3.4 References to component elements

Rendered version is here: https://cellml-specification-dev.readthedocs.io/en/i138_informative_3_4/reference/formal_and_informative/specC04_component_reference.html

Click the "See more" link to (ahem) see more.


src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,105 @@
 
   .. container:: infospec
 
-    **TODO**
\ No newline at end of file
+    .. container:: heading3
+
+      Understanding component references
+
+    As with references to units and variables, the term "component reference" refers to both sides of the naming and calling transaction.
```
MichaelClerx, line 15: What's a "naming and calling transaction" ???

src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,105 @@
...
+    Its use includes both the item which *possesses* the reference (i.e.: a :code:`component` or :code:`import component` with a :code:`name` attribute) as well as the item which is *calling* the reference (i.e.: an item requesting the named component via its own :code:`component_ref` attribute).
```
MichaelClerx, line 16: I don't get this. Surely if X has name "x", and I refer to "x", that referring is not part of X? If I write a book about your book, the reference in my book isn't a property of your book all of a sudden?

MichaelClerx: When I write `<component name="x">` there's no component reference there, right? Just a component and a name?

kerimoyle: So I think we can agree that I have a lot of trouble understanding the "reference" sections in general!  My (latest ...) understanding is that both the name-of-the-thing and the thing-calling-the-name were "references"?  
So for `<component name="myName">` the name attribute value was the reference, and 
for `<connection component_1="myName" ...>` the value of the component_1 attribute was the reference.  This is what I was trying to say with the both sides of a naming/calling bit (though if it's not what's happening, we can ditch that part!).  I've found all of the sections about references really hard to follow ... (which could be just me missing something), but also could show that we need to write them in a much clearer way?  I am happy to be the lowest-understanding-denominator guinea-pig ... but I can't make too many suggestions without actually understanding it first!  Could you give me some uber simple examples perhaps?  Or see whether/where my examples are wrong? Sometimes that helps ... ?

agarny: > So for `<component name="myName">` the name attribute value was the reference, and

?? The `name` attribute value is not a reference. Just the name of the component.

> for `<connection component_1="myName" ...>` the value of the component_1 attribute was the reference.

Here, I agree, the `component_1` attribute value is indeed a reference to the component which name is `myName`.

nickerso: I suspect the confusion might be trying to group `component` and `import component` together. A `component` is a thing. Whereas an `import component` is simultaneously both a `component` thing and a reference to a different thing depending on the context.

kerimoyle: Done in dc25db5

src/reference/informative/informC04_component_reference.rst
kerimoyle: Changed all wording in dc25db5

kerimoyle:
Ready for review please.

src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,189 @@
 
   .. container:: infospec
 
-    **TODO**
+    .. container:: heading3
+
+      Understanding component references
+
+    As with references to units and variables, the term "component reference" refers to places where you need to specify a component by name.
+    This is done through the :code:`component_ref` attribute found in :code:`encapsulation` and :code:`import component` elements, as well as the :code:`component_1` and :code:`component_2` attributes of :code:`connection` elements.
```
agarny, line 16: I would probably replace
```
:code:`import component`
```
with
```
:code:`component` (within an :code:`import` element)
```

Also, I would mention
```
:code:`encapsulation`
```
after (simply because I tend to list imports before encapsulations in a CellML file), i.e. have the following in the end:
```
This is done through the :code:`component_ref` attribute found in :code:`component` (within an :code:`import` element) and :code:`encapsulation` elements, as well as the :code:`component_1` and :code:`component_2` attributes of :code:`connection` elements.
```

src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,189 @@
...
+
+    The complicated part occurs during imports, and involves understanding the *scope* of a component's reference.
+    This is really just the appropriate name by which to call it depending on who is doing the calling.
+
+    Consider the following example.  
+    Here, two families live next door to one another, and within each family unit use the terms "husband" and "wife" to refer to one another.
```
agarny, line 22: What do you mean by "... each family unit use..."?

src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,189 @@
...
+                                        ╵ ╵ ╵ ╵
+      # In LeadbetterFamily.cellml:     ╵ ╵ ╵ ╵ 
+      model: LeadbetterFamilyModel      ╵ ╵ ╵ ╵
+        ├─ component: husband ╴╴╴╴╴╴╴╴╴╴┘ ╵ ╵ ╵
+        └─ component: wife ╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴┘ ╵ ╵
+                                            ╵ ╵
+      # In GoodFamily.cellml:               ╵ ╵
+      model: GoodFamilyModel                ╵ ╵
+        ├─ component: wife ╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴┘ ╵
+        └─ component: husband ╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴┘
```
agarny, line 43: Replace with:
```
model: SurbitonNeighbourhood
  ├─ component: TomGood <───────────────┐
  ├─ component: BarbaraGood <─────────┐ │
  ├─ component: MargotLeadbetter <──┐ │ │
  └─ component: JerryLeadbetter <─┐ │ │ │
                                  │ │ │ │
                           imported components
                                  │ │ │ │
# In LeadbetterFamily.cellml:     │ │ │ │
model: LeadbetterFamilyModel      │ │ │ │
  ├─ component: husband ──────────┘ │ │ │
  └─ component: wife ───────────────┘ │ │
                                      │ │
# In GoodFamily.cellml:               │ │
model: GoodFamilyModel                │ │
  ├─ component: wife ─────────────────┘ │
  └─ component: husband ────────────────┘
```

With the correct font, it will look fine:

Screen Shot 2020-04-08 at 6 23 02 PM

Hmm, I am getting the feeling that you wanted to use some kind of dashed lines to show the imports? I guess I would be happy with either plain or dashed lines. It's just that the dashed lines looked a bit strange to me when I first saw them.

src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,189 @@
...
+          <component name="wife" />
+        </model>
+
+        <!-- Inside the LeadbetterFamily.cellml file: -->
+        <model name="TheLeadbetterFamilyUnit">
+          <component name="husband" />
+          <component name="wife" />
+        </model>
+
+    Now let's make things more interesting by adding an encapsulation hierarchy:
```
agarny, line 88: Before this, I would probably have a sentence or two about the fact that, within `LeadbetterFamilyModel ` and `GoodFamilyModel `, both `husband` and `wife` are component names while, within `SurbitonNeighbourhood`, they are component references (and `TomGood`, `BarbaraGood`, `MargotLeadbetter` and `JerryLeadbetter` are component names).

src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,189 @@
...
+                                            ╵ ╵ ╵ ╵
+      # In LeadbetterFamily.cellml:         ╵ ╵ ╵ ╵
+      model: LeadbetterFamilyModel          ╵ ╵ ╵ ╵
+        ├─ component: husband ╴╴╴╴╴╴╴╴╴╴╴╴╴╴┘ ╵ ╵ ╵
+        └─ component: wife ╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴┼╴┘ ╵
+                                              ╵   ╵
+      # In GoodFamily.cellml:                 ╵   ╵
+      model: GoodFamilyModel                  ╵   ╵
+        ├─ component: husband ╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴┘   ╵
+        └─ component: wife ╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴┘
```
agarny, line 110: In case we decide to use plain lines:

```
model: SurbitonNeighbourhood
  ├─ component: WomenInTheNeighbourhood
  │   ├─ component: BarbaraGood <───────────┐
  │   └─ component: MargotLeadbetter <────┐ │
  └─ component: MenInTheNeighbourhood     │ │
      ├─ component: TomGood <───────────┐ │ │
      └─ component: JerryLeadbetter <─┐ │ │ │
                                      │ │ │ │
                              imported components
                                      │ │ │ │
# In LeadbetterFamily.cellml:         │ │ │ │
model: LeadbetterFamilyModel          │ │ │ │
  ├─ component: husband ──────────────┘ │ │ │
  └─ component: wife ───────────────────┼─┘ │
                                        │   │
# In GoodFamily.cellml:                 │   │
model: GoodFamilyModel                  │   │
  ├─ component: husband ────────────────┘   │
  └─ component: wife ───────────────────────┘
```

src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,189 @@
...
+              <component_ref component="MargotLeadbetter" />
+            </component_ref>
+            <component_ref component="MenInTheNeighbourhood">
+              <component_ref component="TomGood" />
+              <component_ref component="JerryLeadbetter" />
+            </component_ref>
+          </encapsulation>
+        </model>
+
+    This particular encapsulation structure means that the women (Barbara and Margot) are essentially unable to have any contact with the men (Tom and Jerry) even though their original components in the models from which they were imported *were* able to access one another.
```
agarny, line 140: Interesting! This is clearly valid CellML, but I wonder whether this couldn't result in a non-sensical model. I mean, what if `TomGood` and `BarbaraGood` have connected variables in `GoodFamily.cellml`?

src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,189 @@
...
+
+    Note also that these locality naming-calling rules are be applied through multiple generations of importing.
```
agarny, line 142: "... are [to] be applied..."?

src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,189 @@
...
+        ╷ ╷ ╷ ┌╴> └─ component: wife ╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴╴┘
+        ╷ ╷ ╷ ╷
+      imported components
+        ╵ ╵ ╵ ╵
+        ╵ ╵ ╵ ╵  # In CastOfCharacters.cellml:
+        ╵ ╵ ╵ ╵  model: 
+        ╵ ╵ ╵ └╴╴╴├─ component: FelicityKendal
+        ╵ ╵ └╴╴╴╴╴├─ component: RichardBriers
+        ╵ └╴╴╴╴╴╴╴├─ component: PenelopeKeith
+        └╴╴╴╴╴╴╴╴╴└─ component: PaulEddington
```
agarny, line 175: In case we decide to use plain lines:

```
  model: SurbitonNeighbourhood
    ├─ component: WomenInTheNeighbourhood
    │   ├─ component: BarbaraGood <─────────────┐
    │   └─ component: MargotLeadbetter <──────┐ │
    └─ component: MenInTheNeighbourhood       │ │
        ├─ component: TomGood <─────────────┐ │ │
        └─ component: JerryLeadbetter <───┐ │ │ │
                                          │ │ │ │
                                imported components
                                          │ │ │ │
          # In LeadbetterFamily.cellml:   │ │ │ │
          model: TheLeadbetterFamilyUnit  │ │ │ │
  ┌────────> ├─ component: husband ───────┘ │ │ │
  │ ┌──────> └─ component: wife ────────────┼─┘ │
  │ │                                       │   │
  │ │     # In GoodFamily.cellml:           │   │
  │ │     model: TheGoodFamilyUnit          │   │
  │ │ ┌───> ├─ component: husband ──────────┘   │
  │ │ │ ┌─> └─ component: wife ─────────────────┘
  │ │ │ │
imported components
  │ │ │ │
  │ │ │ │  # In CastOfCharacters.cellml:
  │ │ │ │  model:
  │ │ │ └── ├─ component: FelicityKendal
  │ │ └──── ├─ component: RichardBriers
  │ └────── ├─ component: PenelopeKeith
  └──────── └─ component: PaulEddington
```

src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,189 @@
...
+
+    .. code-block:: xml
```
agarny, line 177: Shouldn't we have the following before (i.e. the same as for the other CellML syntax bits):

```
    .. container:: toggle

      .. container:: header

        See CellML syntax
```

src/reference/informative/informC04_component_reference.rst
kerimoyle: I guess I could remove the "unit"?  Would that help it make more sense to you?  It means that within the family of LeadbetterFamily or GoodFamily, the variables are referred to as "husband" and "wife".

src/reference/informative/informC04_component_reference.rst
kerimoyle: Yes, the dashed lines are all imports, the solids are all not.

src/reference/informative/informC04_component_reference.rst
kerimoyle: You tell me ... :) ?  I don't think that connections and mapped variables are imported too? (otherwise the entire imported model would probably come along for the ride ...) And in any case, the two components are totally separate in the importing model, so I'd have thought that no automatic connection was possible ...?

src/reference/informative/informC04_component_reference.rst
kerimoyle: Thanks!

src/reference/informative/informC04_component_reference.rst
agarny: Yes, it would make sense to me if "unit" was removed (and an "s" was added to "use").

src/reference/informative/informC04_component_reference.rst
kerimoyle: I can do a global search and replace if we decide to ditch the dashed ones.  There were several options (including different weights), but more important to get the content in for now, I think.

src/reference/informative/informC04_component_reference.rst
agarny: That's fair enough.

src/reference/informative/informC04_component_reference.rst
kerimoyle: I haven't done the toggling for where there are just short sections of CellML, or where the syntax itself is the important thing.  I guess I figured this was "short enough" ... no strong feelings about it though, I can add if you'd prefer.

src/reference/informative/informC04_component_reference.rst
agarny: The only person who has ever done any advanced things with imports is @nickerso, so I will leave it to him to comment on this... :)

src/reference/informative/informC04_component_reference.rst
agarny: Agreed.

src/reference/informative/informC04_component_reference.rst
agarny: For consistency, I would add a way to toggle the code. It looked "strange" to me to see some code here while earlier I had to toggle similar bits of code.

src/reference/informative/informC04_component_reference.rst
kerimoyle: I'll switch the order, but leave the `import component` as this is the name of an entire section, so want to keep it consistent with titles elsewhere. Done in bd6d599

src/reference/informative/informC04_component_reference.rst
kerimoyle: Done in bd6d599

src/reference/informative/informC04_component_reference.rst
kerimoyle: I'm going to keep the distinction between imports/not-imports for now.  The diagrams get a lot more complicated later on!

src/reference/informative/informC04_component_reference.rst
kerimoyle: Done in bd6d599

src/reference/informative/informC04_component_reference.rst
kerimoyle: Done in bd6d599.

src/reference/informative/informC04_component_reference.rst
kerimoyle: Done in bd6d599

src/reference/informative/informC04_component_reference.rst
kerimoyle: I think I had a conversation with @hsorby about this ... can you confirm what the expected behaviour is, Hugh?

src/reference/informative/informC04_component_reference.rst
kerimoyle: (actually, slightly different tweaking before I read your message, but I think it's clearer now).

src/reference/informative/informC04_component_reference.rst
agarny: I find it confusing. When I first read it, I was "oh, is there something called `import component`?" and then searched for it and sure enough I couldn't find any occurrence of `import component`.

kerimoyle: It's a whole section?? https://cellml-specification-dev.readthedocs.io/en/i138_informative_3_4/reference/formal_and_informative/specB04.html

agarny: No, no, I was talking about me understanding `import component` as being a CellML element of sorts, which it clearly isn't.

kerimoyle: Yeah, I found that idea quite confusing when I first read the spec.  Now I think we should retain the language of `import component` (and `import units`) but reinforce throughout the informative spec that this is a shorthand for component-that-has-been-imported.

agarny: I would be happy if there was to be such a reinforcement. Actually, why not `import➤component` or something similar to distinguish the two elements?

kerimoyle: I think it's better to leave it as the element has been defined in section 2: two separate words, keep it simple.  Especially as the contents of the `import` is not a `component` really, it's a reference to one, so they're not strictly the same thing in any case.  If anything we should change it (everywhere) to be `imported component`, so that the language fits better, but I suspect there would be objections to that too!

agarny: - In CellML/XML, the contents of an `import` element is a `component` element.
- `imported component` is not right either. It is rendered as some code, so I would read `imported` as being the name of a CellML element, which it isn't.

Anyway, I have voiced my view and will leave it to the others to chip in if they want.

src/reference/informative/informC04_component_reference.rst
hsorby: Only connections internal to a component hierarchy are kept therefore any external connections are lost.

agarny: This makes sense, and my comment still stands in that we could end up with something non-sensical, but it's "fine".

kerimoyle: OK, thanks both.  I'll add a comment about that.

src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,197 @@
 
   .. container:: infospec
 
-    **TODO**
+    .. container:: heading3
+
+      Understanding component references
+
+    As with references to units and variables, the term "component reference" refers to places where you need to specify or reference a component using its name (but not assign it its name).
```
MichaelClerx, line 15: I'd drop "specify or" and "(but not assign its name)"

Maybe "reference an existing component by its name" ?

agarny: Agree with @MichaelClerx's remark and suggestion.

agarny: I would also remove "(but not assign it its name)" as this wouldn't be needed anymore if we were to go with @MichaelClerx's suggestion.

kerimoyle: Fixed in ddfc79d

src/reference/informative/informC04_component_reference.rst
```diff
@@ -8,4 +8,197 @@
 
   .. container:: infospec
 
-    **TODO**
+    .. container:: heading3
+
+      Understanding component references
+
+    As with references to units and variables, the term "component reference" refers to places where you need to  refer to a component using its name.
+    This is done through the :code:`component_ref` attribute found in :code:`import component` and :code:`encapsulation` elements, as well as the :code:`component_1` and :code:`component_2` attributes of :code:`connection` elements.
```
agarny, line 16: We don't have something similar for the informative units reference documentation. So, should this be dropped or should something similar be added to the informative units reference documentation? I think I would like the same to be done for units references (after the reordering).

kerimoyle: I think we do, but it will be in another PR, so in another branch of the docs?  See: https://cellml-specification-dev.readthedocs.io/en/i137_inform3_2/reference/formal_and_informative/specC02_units_reference.html

PR #156, opened 2020-04-03
Tidy up 3.4, issue 149

### LINKED ISSUE #149: 3.4.2 confusing wording?
3.4.2 seems odd to me:
- is there ever a component reference which is _not_ a descendant of a model element? If not, why is this phrase there?
- should it perhaps be "direct descendant" instead?

image

nickerso:
Fixes #149, tidying up the rules. The intro paragraph is explicit in the elements and attributes being discussed so no longer a need for the catch-all clause in the rules.


nickerso:
(I know I shouldn't be creating branches in the main repo, but so much easier to just edit the files on github...)

PR #157, opened 2020-04-03
Add link to comment infoitems

### LINKED ISSUE #129: Should we define comments in the spec?

nickerso:
Fixes #129.

### DIFF
--- a/src/reference/sectionA_definitions.inc
+++ b/src/reference/sectionA_definitions.inc
@@ -129,12 +129,12 @@ Specific information items
 
 .. _specA_semantic_equivalence:
 
-Semantically equivalent CellML infosets
----------------------------------------
+Non-specific information items
+------------------------------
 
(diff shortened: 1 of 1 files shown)


src/reference/sectionA_definitions.inc
```diff
@@ -134,7 +134,7 @@ Semantically equivalent CellML infosets
 
 #. Two :ref:`CellML infosets<specA_cellml_infoset>` SHALL be deemed semantically equivalent if one can be transformed into the other by making any or none of the following changes:
 
-   #. Adding, removing, and/or modifying comment information items.
+   #. Adding, removing, and/or modifying comment information items (https://www.w3.org/TR/xml-infoset/#infoitem.comment).
```
agarny, line 137: I would make "comment information items" a link.

nickerso: good idea, done that.

PR #166, opened 2020-04-08
Single page generation suitable for cellml.org

hsorby:
To actually generate the **official** page the ``unofficial`` flag must be set to ``False``.  The contents of the generated ``index.html`` file can then be directly copied and pasted into the plone CMS as the HTML content of that page.


MichaelClerx:
Not sure how I review this :D

hsorby:
This PR request as the official/unofficial switch in it so people building the specification will get a `this is an unofficial build of the documentation` in the generated docs.  Mitigating people thinking that they are looking at the official specification.

agarny:
I have just generated an HTML version of the spec using this PR, but... where can I see that it's an unofficial version of the spec. It all looks fine to me...?

hsorby:
Under the heading for the specification.

MichaelClerx:
Maybe we can update `release = '2.0.0-rc1'` to include something about it not being the official release on every page?

hsorby:
Can do.

agarny:
Screenshot please? I just can't see it...

FWIW, I have done `make singlehtml` and then opened `build/singlehtml/index.html`, but I cannot see `unofficial` anywhere.

Also, why for the table of contents, we can only see one level of sections (i.e. `1. Definitions`, but not `1.1. Terminology` for example)? I find that second level very useful to quickly access a particular subsection.

MichaelClerx:
> Also, why for the table of contents, we can only see one level of sections (i.e. `1. Definitions`, but not `1.1. Terminology` for example)? I find that second level very useful to quickly access a particular subsection.

New issue :D

MichaelClerx:
> FWIW, I have done `make singlehtml` and then opened `build/singlehtml/index.html`, but I cannot see `unofficial` anywhere.

Screenshot from 2020-04-17 01-03-46

It's well hidden!

By adding it to the `release` it should show up in other places as well

MichaelClerx:
Also says "CellML 2.0" (`project` name) and then `2.0.0-rc1` (`release`), which seems a bit too much?

agarny:
> > Also, why for the table of contents, we can only see one level of sections (i.e. `1. Definitions`, but not `1.1. Terminology` for example)? I find that second level very useful to quickly access a particular subsection.
> 
> New issue :D

See issue #286.

agarny:
> > FWIW, I have done `make singlehtml` and then opened `build/singlehtml/index.html`, but I cannot see `unofficial` anywhere.
> 
> Screenshot from 2020-04-17 01-03-46
> 
> It's well hidden!
> 
> By adding it to the `release` it should show up in other places as well

How did you build the documentation exactly? I am building it directly off @hsorby's `single_page_for_cellml_org` branch, which is not up-to-date with the `cellml-2-drafting`. Maybe that's the reason? It's a mess and I really feel like I am wasting my time here.

MichaelClerx:
Good point! I'm actually building the old version

hsorby:
I have updated this PR to the latest from `cellml-2-drafting`.

agarny:
Great, I can *now* see it:

Screen Shot 2020-04-17 at 12 26 15 PM

I thought for a second that I was becoming crazy (sill possible though!)!

agarny:
Ok, now that I see what is visible for an unofficial version of the spec, would it actually be possible to have some kind of ribbon indicating that this is an unofficial version of the spec? Somethink like the ["Fork me" ribbon](https://github.blog/2008-12-19-github-ribbons/)?

I guess the current solution is better than nothing, just that it could easily be missed.

MichaelClerx:
> Ok, now that I see what is visible for an unofficial version of the spec, would it actually be possible to have some kind of ribbon indicating that this is an unofficial version of the spec? Somethink like the ["Fork me" ribbon](https://github.blog/2008-12-19-github-ribbons/)?
> 
> I guess the current solution is better than nothing, just that it could easily be missed.

There's `html_css_files = ['css/cellml.css']` in `conf.py`. I guess we could add an extra file there in unofficial mode?

hsorby:
We can add some content to the html template and `if` it in and out.

agarny:
Sounds like a plan then! :)

kerimoyle:
Hang on ... isn't the single-page HTML meant to be the official version?? It's hosted on our site, identical to the PDF??

MichaelClerx:
It's just for the version people generate locally

hsorby:
The HTML won't be identical to the PDF that's impossible.  We can only say that it is generated from the same content.

kerimoyle:
That's what I mean ... without all the informative bits through it.

hsorby:
Right try building the html version now.  It should have a ribbon across the top right hand corner of every page saying 'unofficial version'.  Try and not notice that @agarny!

hsorby:
Do we need/want something similar for PDF generation?

hsorby:
Or do we let @MichaelClerx get some sleep?

agarny:
> Right try building the html version now. It should have a ribbon across the top right hand corner of every page saying 'unofficial version'. Try and not notice that @agarny!

Woo hoo! :) (I am not super keen on the greenish side of it, but... meh! :))

agarny:
> Do we need/want something similar for PDF generation?

If possible, it would be good to have it there too.

agarny:
Incidentally, this means that we don't that "This is not the official version of the normative specification." sentence anymore...?

agarny:
> Or do we let @MichaelClerx get some sleep?

He's young, he doesn't need to sleep. :)

hsorby:
> Incidentally, this means that we don't that "This is not the official version of the normative specification." sentence anymore...?

We could do away with that if the PDF version can be updated to have some unofficial version added in somewhere.

agarny:
> > Incidentally, this means that we don't that "This is not the official version of the normative specification." sentence anymore...?
> 
> We could do away with that if the PDF version can be updated to have some unofficial version added in somewhere.

Waiting then. :)

hsorby:
Actually the generated version of the PDF document has this in the header of every non-chapter page:

image

hsorby:
That is the version number will have `unofficial` appended to it.

agarny:
I guess it's fine then. Cool!

kerimoyle:
Ready to go?

hsorby:
Do we want the `this is not an official ... ` part taken out of the documentation?

kerimoyle:
Only if we have something which also tells them why, and points to the place which is official?

hsorby:
I think we will go as is then.

kerimoyle:
Ooops, I meant the other way around.  If we have it, I think we should point to the "official" PDF version?

kerimoyle:
So: 

"This is not the official version of the normative specification. You can find that at ???"

hsorby:
It doesn't as it is given a few lines down it was added with the sole purpose of saying this is not an official version.

kerimoyle:
Ah, yep, sorry.  Can we add "official" to that wording below?

hsorby:
> Can we add "official" to that wording below?

Just added this as an item on final tweaks.

Issue #173, opened 2020-04-09
3.10.9.2-3: rewording??


kerimoyle:
Currently we have this: (see https://cellml-specification-dev.readthedocs.io/en/i143_3_10_map_variables_informative/reference/formal_and_informative/specC10_interpretation_of_map_variables.html)

> 9. The applicable interfaces for variables A and B SHALL be defined as follows.
> 9.1 When the parent component element of variable A is in the sibling set of the parent component element of variable B, the applicable interface for both variables A and B SHALL be the public interface.
> 9.2 When the parent component element of variable A is in the encapsulated set of the parent component element of variable B, the applicable interface for variable A SHALL be the public interface, and the applicable interface for variable B SHALL be the private interface.
> 9.3 When the parent component element of variable B is in the encapsulated set of the parent component element of variable A, the applicable interface for variable A SHALL be the private interface, and the applicable interface for variable B SHALL be the public interface.

a) I also think that the language here is really dense, and we could do with defining a component AA and component BB (instead of having to say "the parent component element of variable A" etc) to make it easier to explain?
b) It's been explained (I think) to me before, but reading it again I still feel that points 2 and 3 are identical??

Here's my suggestion for rewording:

1. When component AA and BB are sibling components, the applicable interface for all variables in both components SHALL be the public interface.
2. When component AA is a parent of component BB, the applicable interface value for all variables in AA SHALL be the private interface, and the applicable interface for all variables in BB SHALL be the public interface.

MichaelClerx:
I would agree with that. We really don't need all these sibling sets etc. just to make this point!

- same parent? public-public
- parent-child? public-private
- anything else? can't be connected

Issue #172, opened 2020-04-09
3.10 order of points: move 6 and 7 to the end?


kerimoyle:
Can we move the current points 3.10.6-7 to the end of the list?  This means that we address what's needed to create a connection (the interface stuff) before getting on to units stuff?

agarny:
I would be ok with that.

MichaelClerx:
Don't see why not

Issue #171, opened 2020-04-08
3.10 language: remove 3.10.4?


kerimoyle:
3.10.4: CellML models MUST NOT contain any pair of map_variables elements which duplicates an existing arc in the variable equivalence network.
3.10.5: The variable equivalence network MUST NOT contain any cycles.

If we have point 5, does point 4 become redundant? A - B - A is a cycle?

kerimoyle:
... and when it comes down to it, why are we not allowed cycles in the first place??  Surely it should be more of a matter of having a definition somewhere in the set/network (just one, not more) as for any other variable?  Why do we care how it got there?

MichaelClerx:
Agreed with the idea that 3.10.4 is redundant (since edges don't have a direction)

I like the rule of not allowing cycles though. It makes it easier to work with a (n already validated) CellML model, you can traverse connections without having to check for cycles.

It also means you can't create an infinite number of models equivalent to a given model by just re-adding the same `<map_variables>` elements over and over again.

Finally I just think it's pleasing! The sparsest graph that does what we want it to, has no cycles. Adding cycles just creates redundant extra flappy bits of CellML that I'd rather be without

MichaelClerx:
use different terminology to describe the "variable equivalence network"?
Pros of "network

Issue #170, opened 2020-04-08
3.10 language: network vs set?


kerimoyle:
Should we use different terminology to describe the "variable equivalence network"?  
Pros of "network": it implies connectivity requirements, which makes the language around interfaces and connections more straightforward;
Cons of "network": it's not really true, it's more of a set (in the end)?

MichaelClerx:
I thought the idea was that the network is for all variables in the model? So a graph where each node is a variable, and then and edge is added for each connection, and finally starting from variable X, any variable Y you can reach by travelling over the edges is part of X's connected variable set?

MichaelClerx:
I agree this is a bit cumbersome :D

kerimoyle:
Agreed -  I think.  So there is a _global_ variable _network_ (connected by maths) but _equivalent_ variable _sets_ (connected by mapping)?

kerimoyle:
The second sentence of 1) doesn't fit with that though ...

> For the purposes of this specification, the variable equivalence (conceptual) network SHALL be an undirected graph with one node for every variable element in the CellML model. **The arcs of this graph SHALL be equivalences defined in the CellML model.**

Oh dear, I've confused myself again.  Sigh.

kerimoyle:
And the last point says:

> The `variable` elements in a CellML model SHALL be treated as belonging to a single "connected variable set".  Each set of connected variables is the set of all `variable` elements for which the corresponding nodes in the variable equivalence network form a connected subgraph.  Each set of connected variables represents one variable in the underlying mathematical model.

I don't understand the first sentence?  We start by saying there is a single connected variable set, then in the next sentence say that there are lots of them ...?  I know that all the variables are connected through the maths, but we haven't mentioned that anywhere here ... Can we change the first sentence so that either specifies the maths connection explicitly, or at least doesn't imply that they're all "connected" via the same mechanism?

MichaelClerx:
> The second sentence of 1) doesn't fit with that though ...
> 
> > For the purposes of this specification, the variable equivalence (conceptual) network SHALL be an undirected graph with one node for every variable element in the CellML model. **The arcs of this graph SHALL be equivalences defined in the CellML model.**
> 
> Oh dear, I've confused myself again. Sigh.

Good catch. I think "equivalences" should be connections there. We have an arc from x1 to x2 if they are directly connected, but not an arc from x1 to every variable in its connected variable set!

MichaelClerx:
> And the last point says:
> 
[...]

PR #175, opened 2020-04-09
Rewording points in 3.10

kerimoyle:
Addresses #170, #171, #172, #173 

This is a straw-man for some rewording of the map variables section in chapter 3.


src/reference/sectionC_interpretation.inc
```diff
@@ -373,19 +367,25 @@ Interpretation of ``map_variables`` elements
...
 
-#.  The :code:`variable` elements in a CellML model SHALL be treated as belonging to a single "connected variable set".
-    Each set of connected variables is the set of all :code:`variable` elements for which the corresponding nodes in the variable equivalence network form a connected subgraph.
-    Each set of connected variables represents one variable in the underlying mathematical model.
+#.  For each :code:`map_variables` element present in the CellML model, the variable unit reduction (see :ref:`Effect of units on variables <specC_effect_of_units_on_variables>` ) of variable *A* MUST have an identical set of tuples to the variable unit reduction of variable *B*.
+    Two sets of tuples SHALL be considered identical if all of the tuples from each set are present in the other, or if both sets are empty.
+    Two tuples are considered identical if and only if both the name and exponent value of each tuple are equivalent.
+
+#.  Tuples differing by a multiplying factor in their unit reduction MUST be taken into account when interpreting the numerical values of the variables (see :ref:`Interpretation of units<specC_interpretation_of_units>`).
```
MichaelClerx, line 384: Is this line not better left for the interpretation section?

src/reference/sectionC_interpretation.inc
```diff
@@ -373,19 +367,25 @@ Interpretation of ``map_variables`` elements
...
+
+#.  The :code:`variable` elements in a CellML model SHALL be treated as belonging to a set, connected by the relationships specified in the :code:`math` elements of the model.
```
MichaelClerx, line 386: Again, this bit goes in the interpretation part

MichaelClerx:
If we go with this (which I'd like), a lot of 3.9 can go. Who needs all these sets!

kerimoyle:
Or maybe just a simple glossary at the beginning of chapter 3? then we can define the element interpretation in a sensible order independent of definitions?

MichaelClerx:
Hmmmm! No I like the idea of the interpretation saying "this is what the syntax means", so not quite a glossary, just some actual rules about what encapsulation _means_ ?

kerimoyle:
Haha, no I mean a glossary that explains some convenient shorthand for the interpretation section, eg: test value = value of the variable referenced by the reset element’s `test_variable` attribute, etc

kerimoyle:
Though what encapsulation (or anything) means should (haven't checked ...) be explained in the informative spec as an opener?

MichaelClerx:
For the avoidance of doubt, what we currently have is correct so I'm not too upset if it stays in, but I think it could be a whole lot simpler, e.g.

> - Connections can only be made between variables of components that are either siblings in the encapsulation hierarchy, or have a parent-child relationship.
> - Variables of components that are siblings in the encapsulation hierarchy are connected through the public interface.
> - In parent-child relationships, the variable in the parent component uses the private interface, while the variable in the child component uses the public interface.

kerimoyle:
Changes to 3.10.9 as per meeting: https://cellml-specification-dev.readthedocs.io/en/normativerewording_3_10/reference/formal_and_informative/specC10_interpretation_of_map_variables.html

kerimoyle:
This still needs the informative sections to be reordered to match the change in order of 3.10, but I'll sort them out separately.

Issue #136, opened 2020-04-01
Informative 3.1 Interpretation of import elements


kerimoyle:
Write the "see more" block for this section.

PR #177, opened 2020-04-09
i136 Interpretation of imports, informative

kerimoyle:
Teeny weeny PR, just adding a redirect to another section in the See More block.

Addresses #136


src/reference/informative/informC01_imports.rst
```diff
@@ -8,8 +8,4 @@
 
   .. container:: infospec
 
-    CellML 2.0 is not xenophobic.
-    When an element (either a :code:`component` or a :code:`units` item) is imported, it is treated in the same way as the locally defined items.
-
-
-    **TODO** Don't fully understand the first sentence...
+    Please see the informative sections on :ref:`Units reference<specC_units_reference>` and :ref:`Component reference<specC_component_reference>` pages for examples specific to importing :code:`units` and :code:`component` elements.
```
agarny, line 11: Very minor: add a space between "reference" and "<specC_..."? I believe I have seen you correcting code here and there to do just that?

kerimoyle: There are two different formats around whitespace depending on whether it's an internal (via the `:ref:` role) or external (via the named links roles, eg `:mathml:` or `:unicode:`).  Internal ones must have no whitespace before the <> but external ones must have a space.  Here they are internal references, so no space?

agarny: I honestly don't know. Have you tried with and without space? Anyway, I have approved, so if needed this can be "fixed" later.

kerimoyle: OK, ta.  That's the formatting that I've been following in any case.  It will need to be changed everywhere later if deemed "wrong".

PR #191, opened 2020-04-10
Slight rewording of 3.11 resets

MichaelClerx:


src/reference/sectionC_interpretation.inc
```diff
@@ -401,8 +401,8 @@ Interpretation of ``reset`` elements
 #. All :code:`reset` elements SHALL be considered sequentially for the connected variable set (see :ref:`Interpretation of map_variables<specC_interpretation_of_map_variables>`) to which the referenced variable belongs.
    The sequence SHALL be determined by the value of the reset element’s :code:`order` attribute, lowest (least positive / most negative) having priority.
 
-#. The condition under which a reset occurs SHALL be defined by the equality of the reset element’s :code:`test_variable` attribute and the evaluation of the MathML expression encoded in the :code:`test_value`.
+#. The condition under which a reset occurs SHALL be defined by the equality of value of the variable referenced by the reset element’s :code:`test_variable` attribute and the evaluation of the MathML expression encoded in the :code:`test_value`.
```
kerimoyle, line 404: Could we first define "reset value" and "test value" and "test variable" as terms that mean "value of the variable referenced by the xx attribute"? So that these sentences can be actually readable ...?

kerimoyle: also "the value of the variable ..."

MichaelClerx:
Replaced by #215

Issue #183, opened 2020-04-10
Real number strings must include basic real number strings


MichaelClerx:
> 5. A real number string:
>    1. SHALL be a base 10 representation of a real number r = s × 10 e where s is the significand, a real number, and e is the exponent, an integer.
>    2. The representation of the number SHALL be the representation of the significand followed immediately by the representation of the exponent.
>    3. The significand SHALL be represented as a basic real number string.
>    4. An exponent SHALL be represented by an exponent separator character, followed by the integer string representation of the value of the exponent. Non-negative exponents MAY begin with the Basic Latin plus sign character U+002B as the sign indicator. The exponent separator character SHALL be either the Basic Latin “E” character U+0045 or the Basic Latin “e” character U+0065.
>    5. A real number string without an exponent SHALL be a basic real number string.

This bit:

> The representation of the number SHALL be the representation of the significand followed immediately by the representation of the exponent.

says that you _have_ to have a "representation of the exponent".
That would be OK if "" was a valid representation of an exponent, but points 4 and 5 don't seem to allow that.

Two things I'd like to see resolved:

- it needs to be clear that the "real number string" concept includes a "basic real number string". So if we say a "real number string" is required it does _not_ mean you need an "e" or "E" in there. Point 5 seems to say the opposite.
- what about the case "1.2e"?

kerimoyle:
Two things I'd like to see resolved:

> - it needs to be clear that the "real number string" concept includes a "basic real number string". So if we say a "real number string" is required it does not mean you need an "e" or "E" in there. Point 5 seems to say the opposite.

v. "A real number string without an exponent SHALL be a basic real number string." ?

> - what about the case "1.2e"?

Point 5 comes after point 4, which says "exponent SHALL be represented by an exponent separator character, followed by the integer string representation of the value of the exponent."  So 1.2e isn't valid because there is no integer string representation of the value of the exponent (it should be 1 when the number is blank, so would be 1.2e1, which is valid).

MichaelClerx:
Both of these are fixed in #199

Issue #178, opened 2020-04-10
"European numeral"


MichaelClerx:
Not a great name?

i. [Wiki lists two numeral systems referred to by this name](https://en.wikipedia.org/wiki/European_numerals)
ii. They're either hindu or arabic (wiki doesn't agree with itself on this)

Interesting sentence from article above:

> Other alternative names are Western Arabic numerals, Western numerals, Hindu numerals, and Unicode calls them digits.[5]

Given that we refer to the unicode spec., and that the unicode people spend a looooooot of time on international commiitteess, shall we just call them "digits" too?

Issue #33, opened 2020-03-07
No + in number strings?


MichaelClerx:
We're currently allowing `-1.23` and also `1e+23`, but not `+2` or `+1.23e+456`

agarny:
I think it's imtentional. @nickerso?

nickerso:
yep, imtentional :) @hsorby was arguing that some languages include the leading `+` by default, but we can't find any examples in practice.

MichaelClerx:
ok!

MichaelClerx:
Wait, what? In any language I know you can write `x = +1`
Including MathML, for example

And we're allowing `1e+2` too

kerimoyle:
The only place we're allowing the + is in the exponents, yeah.  Not sure on the fundamental reason for that though...

nickerso:
We allow in the exponential because some languages/tools will write it out. There are no examples we can find where a tool or language will write a `+` in front of positive numbers. Technically it can certainly be done but it generally is not done in practice. Are there known examples where this is needed? (As always, we don't stop people putting whatever they want into a MathML number...

kerimoyle:
Does that mean that this is valid?
```
<math>
  <apply><eq/>
      <ci>x</ci>
      <cn>+1</cn>
   </apply>
</math>
```
... but this is not ... ?
```
<variable name="x" initial_value="+1"/>
```
??

nickerso:
yep - CellML doesn't describe the rules for valid MathML, we just say which elements from MathML 2.0 are valid to use in a CellML model.

MichaelClerx:
> We allow in the exponential because some languages/tools will write it out. There are no examples we can find where a tool or language will write a `+` in front of positive numbers. Technically it can certainly be done but it generally is not done in practice. Are there known examples where this is needed? (As always, we don't stop people putting whatever they want into a MathML number...

That's a very tool-based reason!

In my mind sticking a plus or a minus in front of a number are just two opposite equal things, so it seems a bit weird to disallow one, while every tool I know of will _parse_ "+1" just as easily as it will parse "-1"

MichaelClerx:
Also allowing it in the exponent but then not in normal numbers really grates me

nickerso:
Thats what I would have thought @MichaelClerx, but thought your earlier `OK` and closing the issue resolved things :) Thought we had got off lightly on this one!

I'm not strictly against allowing the `+` so happy to add it if people think it would be useful.

[...]

PR #199, opened 2020-04-14
Changes to number string definitions

MichaelClerx:
Normative changes:

- "European numerals" are now called "Digits" #178
- Non-negative integer string must be one or more digits: Not sure if we need to allow a leading plus here? I've gone for "no" in this PR.
- Integer strings start with + or -, then one or more digits
- made clear at the outset that the exponent in real number strings is _optional_ #183 
- tweaked the wording so that "" is not a number string, nor is ".", but "1." and ".1" both are

Informative changes:

- Numerals --> digits #33
- Basic latin --> Basic latin alphabetic characters ("Basic Latin" includes digits, @ etc., all of ascii! https://en.wikipedia.org/wiki/Basic_Latin_(Unicode_block) )
- small tweaks


MichaelClerx:
Closes #199

MichaelClerx:
Closes #183

src/reference/informative/informA1.rst
```diff
@@ -16,35 +16,39 @@
...
-    It contains the upper and lowercase alphabet without decoration:
+    The codes themselves are arranged into blocks representing sets of characters, and the Basic Latin block is one of these.
+    It contains, amongst other things, the upper and lowercase alphabet without decoration:
 
       :code:`abcdefghijklmnopqrstuvwxyz` (symbols between :unicode:`0061` and :unicode:`007A`)
       :code:`ABCDEFGHIJKLMNOPQRSTUVWXYZ` (symbols between :unicode:`0041` and :unicode:`005A`)
 
-    The European numerals are the Unicode set:
+    These characters are refferd to in CellML as the Basic Latin alphabetic characters.
+
```
kerimoyle, line 33: Do we want quotes or italics for "Basic Latin alphabetic characters"?

src/reference/informative/informA1.rst
```diff
@@ -16,35 +16,39 @@
...
+    The digits are the Unicode characters:
 
       :code:`0123456789` (symbols between :unicode:`0030` and :unicode:`0039`)
 
+    Together with the Basic Latin alphabetic characters, they form the Basic Latin alphanumerical characters.
```
kerimoyle, line 38: Is it "alphabetic" and "alphanumeric"  ... or "alphabetical" and "alphanumerical"? There doesn't seem to be any difference in the meaning of each of the pairs.

src/reference/informative/informA1.rst
```diff
@@ -16,35 +16,39 @@
...
-    - the tab (:unicode:`0009`), 
+
+    - the space (:unicode:`0020`),
+    - the tab (:unicode:`0009`),
     - the carriage return (:unicode:`000D`), and
-    - the line feed (:unicode:`000A`).  
+    - the line feed (:unicode:`000A`).
 
-    Together these symbols define the only characters which can be interpreted into values for CellML element attributes and content. 
+    Together these symbols define the only characters which can be interpreted into values for CellML names and numbers.
```
kerimoyle, line 54: It's not quite the same thing?  We can only use these characters anywhere that something needs to be interpreted, including references etc?  Hmmm ... and now thinking about it, what about all the possible symbols in URLs?? We haven't listed them anywhere ...?

src/reference/sectionA_definitions.inc
```diff
@@ -250,15 +250,15 @@ The following data representation formats are defined for use in this specificat
 
    .. container:: issue-data-repr-basic-real-sign
 
-      2. SHALL, when the basic real number being represented is negative, begin with the Basic Latin hyphen-minus character :unicode:`002D` as the sign indicator.
+      2. MAY begin with a single Basic Latin plus character :unicode:`002B` or a Basic Latin hyphen-minus character :unicode:`002D` as the sign indicator.
```
kerimoyle, line 253: When it's negative it must begin with the - sign?  

- MAY, when the basic real number being represented is positive, begin with the Basic Latin plus character :unicode:`whateveritis` as the sign indicator.
- SHALL, when the basic real number being represented is negative, begin with the Basic Latin hyphen-minus character :unicode:`002D` as the sign indicator.

kerimoyle:
Looks pretty good, thanks :) Couple of follow up questions:
- Did you do a global search for the old `specA_european_numeral` link so we're sure it's not referenced elsewhere?  
- Also, it's probably worth checking that any manually added numbers haven't been broken?  Just search for the ":hardcodedref:" tag to see where we've used section numbers in the text.

kerimoyle:
Also would be good to check that none of the informative examples are broken by this change?  Might add that to my list though ...

src/reference/informative/informA1.rst
MichaelClerx: refferd --> referred

src/reference/sectionA_definitions.inc
```diff
@@ -218,7 +218,7 @@ The following data representation formats are defined for use in this specificat
 
    .. container:: issue-data-repr-nneg-int-euro-num
 
-      2. SHALL consist entirely of :ref:`European numerals<specA_european_numeral>`.
+      2. SHALL consist of one or more :ref:`digits<specA_digit>`.
```
MichaelClerx, line 221: The "non-negative integer string" can go, not used in the spec anymore

src/reference/informative/informA1.rst
MichaelClerx: We refer to the xlink spec for URLs

> The value of this attribute SHALL be a valid locator href, as defined in Section 5.4 of the XLink specification.
> The href attribute SHALL be treated according to the XLink specification, by applying the rules for simple-type elements.

MichaelClerx: Have replaced this sentence by something less restrictive. (Just saying these classes will be uesd, not that they will be used exclusively!)

src/reference/sectionA_definitions.inc
MichaelClerx: I'm fine with having positive and negative numbers defined outside the spec, in the usual way. It's captured by the first rule (base 10 int representation) I think

MichaelClerx:
> Looks pretty good, thanks :) Couple of follow up questions:
> 
>     * Did you do a global search for the old `specA_european_numeral` link so we're sure it's not referenced elsewhere?

Yes

MichaelClerx:
> 
>     * Also, it's probably worth checking that any manually added numbers haven't been broken?  Just search for the ":hardcodedref:" tag to see where we've used section numbers in the text.

the result of grepping is

> ./src/conf.py:# more easily by searching for the "hardcodedref" tag.
> ./src/conf.py:.. role:: hardcodedref 

so maybe they're all in PRs ?

MichaelClerx:
Fixed the informative spec now too

MichaelClerx:
Happy to merge now @kerimoyle ?

kerimoyle:
> so maybe they're all in PRs ?

Could be ... though I'm a bit surprised.  Is your branch up to date?

MichaelClerx:
> Are we changing "alphabetic" into "alphabetical" and "alphanumeric" into "alphanumerical" in this PR or somewhere else?

Separate PR

MichaelClerx:
> > so maybe they're all in PRs ?
> 
> Could be ... though I'm a bit surprised. Is your branch up to date?

Ah. Done it better now. Checked and they're ok!

Issue #182, opened 2020-04-10
1.3.1.3 and 1.3.1.4 use informal language for some reason


MichaelClerx:
> 3. SHALL contain at least one alphabetic character.
> 4. SHALL NOT begin with a numeral or an underscore.

why are we saying "alphabetic" all of a sudden when we've defined clearer terms on page 2?

MichaelClerx:
This can be dropped if we do #181 :D

Issue #181, opened 2020-04-10
Identifier rules can be condensed


MichaelClerx:
1.3.1.2. says "only letters, numbers, underscores" and then

1.3.1.3 at least one letter
1.3.1.4 must start with not-a-number-or-underscore

Can we replace 1.3.1.3 and 1.3.1.4 with "must start with a letter" ?

MichaelClerx:
The answer is: No! Because we dropped the rule saying "at least one char" in favour or "at least one char"

MichaelClerx:
1 and 2 are also a bit redundant, first saying "must be unicode chars", then saying "must be a subset of unicode chars"

MichaelClerx:
How about

> 1. A CellML identifier
>     1. SHALL consist of a Basic Latin alphabetic character, followed by any number of Basic Latin alphanumeric characters and/or Basic Latin underscore characters
>     2. SHALL, when comparing...

MichaelClerx:
or maybe clearer:

> 1. A CellML identifier
>     1. SHALL consist of a Basic Latin alphabetic character, which MAY be followed by any combination of Basic Latin alphanumeric characters or Basic Latin underscore characters
>     2. SHALL, when comparing...

PR #200, opened 2020-04-14
Identifier rules

MichaelClerx:
- condensed identifier definition (#181 and #182 )

- the equality rule said "have identical sequences", which I think meant "have sequences that are entirely identical", but could be taken to mean "have identical sequences anywhere in the string", e.g. "hello_me" == "hello_you" since both have an identical sequence of characters ("hello_").
  - Someone might still want to argue that an identifier isn't a sequence itself so that "are" is wrong? Not me!

### DIFF
--- a/src/reference/sectionA_definitions.inc
+++ b/src/reference/sectionA_definitions.inc
@@ -188,23 +188,11 @@ The following data representation formats are defined for use in this specificat
 
    .. container:: issue-data-repr-identifier-unicode
 
-      1. SHALL be a sequence of Unicode characters.
-
-   .. container:: issue-data-repr-identifier-latin-alphanum
-
-      2. SHALL NOT contain any characters except :ref:`Basic Latin alphanumeric characters<specA_basic_latin_alphanumeric_character>` and :ref:`Basic Latin underscores<specA_basic_latin_underscore>`.
-
-   .. container:: issue-data-repr-identifier-at-least-one-alphanum
-
-      3. SHALL contain at least one :ref:`alphabetic<specA_basic_latin_alphabetic_character>` character.
-
-   .. container:: issue-data-repr-identifier-begin-euro-num
-
-      4. SHALL NOT begin with a :ref:`numeral<specA_european_numeral>` or an :ref:`underscore<specA_basic_latin_underscore>`.
+      1. SHALL consist of a single :ref:`Basic Latin alphabetic character<specA_basic_latin_alphabetic_character>`, which MAY be followed by any combination of :ref:`Basic Latin alphanumeric characters<specA_basic_latin_alphanumeric_character>` and :ref:`Basic Latin underscores<specA_basic_latin_underscore>`.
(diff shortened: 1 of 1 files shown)


kerimoyle:
NUMBER 200! Wahey!!

MichaelClerx:
This also closes #182

src/reference/sectionA_definitions.inc
```diff
@@ -188,23 +188,11 @@ The following data representation formats are defined for use in this specificat
...
-      2. SHALL NOT contain any characters except :ref:`Basic Latin alphanumeric characters<specA_basic_latin_alphanumeric_character>` and :ref:`Basic Latin underscores<specA_basic_latin_underscore>`.
-
-   .. container:: issue-data-repr-identifier-at-least-one-alphanum
-
-      3. SHALL contain at least one :ref:`alphabetic<specA_basic_latin_alphabetic_character>` character.
-
-   .. container:: issue-data-repr-identifier-begin-euro-num
-
-      4. SHALL NOT begin with a :ref:`numeral<specA_european_numeral>` or an :ref:`underscore<specA_basic_latin_underscore>`.
+      1. SHALL consist of a single :ref:`Basic Latin alphabetic character<specA_basic_latin_alphabetic_character>`, which MAY be followed by any combination of :ref:`Basic Latin alphanumeric characters<specA_basic_latin_alphanumeric_character>` and :ref:`Basic Latin underscores<specA_basic_latin_underscore>`.
```
kerimoyle, line 191: Perhaps: "SHALL **begin** with a single ... which MAY be followed "... etc?  It doesn't change the meaning, but it makes it easier (I think ...) to get the correct idea the first time you read it?  

Also, maybe "Latin alphanumeric characters **and/or** Basic Latin underscores" ?

src/reference/sectionA_definitions.inc
```diff
@@ -188,23 +188,11 @@ The following data representation formats are defined for use in this specificat
...
 
    .. container:: issue-data-repr-identifier-identical
 
-      5. SHALL, when comparing two identifiers, be considered identical to another identifier if and only if both identifiers have identical sequences of characters.
+      2. SHALL, when comparing two identifiers, be considered identical to another identifier if and only if both identifiers are identical sequences of characters.
```
kerimoyle, line 195: Perhaps: "if and only if both identifiers consist of the same characters in the same order." or, if we're being fancy "identical characters in an identical sequence."?  The current definition feels (it isn't, but it comes across as) a wee bit circular...?

src/reference/sectionA_definitions.inc
MichaelClerx: I'm gonna stick with "consist of" because if we say "begins with X, may be followed by Y", then we're not explicitly saying it isn't then followed by P,Q,R and/or S

MichaelClerx:
Approved in meeting, so merging

Issue #179, opened 2020-04-10
Hyperlink for comment information items in 1.2.3


MichaelClerx:
I remember the discussion that led to adding it, but we don't have a hyperlink for the other 10 info item types, so don't really see why this one should have one?

MichaelClerx:
Maybe we should just start with a line like

> For the purposes of this specification, a specific information item is one of the following (see https://www.w3.org/TR/xml-infoset/#infoitem for definitions):

again, and then list the 3 non-specific ones?

PR #202, opened 2020-04-14
Non-specific item section

### LINKED ISSUE #180: Why is "Character information items" its own section?
Currently it's 1.2.4, but I don't see why it shouldn't be part of 1.2.3

MichaelClerx:
- Merged 1.2.3 and 1.2.4 (#180)
- Added definition of non-specific items w. link (#179)

Also made a slight change to the link which you may or may not like...


src/reference/sectionA_definitions.inc
```diff
@@ -105,7 +105,7 @@ CellML and XML
 Specific information items
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-#. For the purposes of this specification, a specific information item is one of the following (see https://www.w3.org/TR/xml-infoset/#infoitem for definitions):
+#. For the purposes of this specification, a specific information item is one of the following (see https://www.w3.org/TR/xml-infoset/ for definitions):
```
kerimoyle, line 109: I should have done this already ... but it would be good to have this included in the external links section in the config file?  Makes it easier to see what we're using if they're all listed there?  We can still write out the address here though.

kerimoyle: We have the :xml_infoset: tag already (see below).  Should this be something less specific??  It's currently only used at the beginning of the Terminology section (second sentence of entire doc):
'xml_infoset': ('http://www.w3.org/TR/2004/REC-xml-infoset-20040204/%s', ''),

nickerso: pretty sure this change was just to link to the top of the page as the table of contents gives a better "overview" of the different types of info items. The other links are to specific items, so the more specific link is correct.

src/reference/sectionA_definitions.inc
```diff
@@ -146,13 +156,6 @@ Non-specific information items
       changing the number of whitespace characters in such an information item,
       or changing the number of whitespace characters at the beginning or end of any character information item.
 
-.. marker_cellml_information_sets_3
```
nickerso, line 149: removing this marker will need a change elsewhere to match right? or is this one not used directly @kerimoyle ?

src/reference/sectionA_definitions.inc
MichaelClerx: Making separate issue for this!

src/reference/sectionA_definitions.inc
MichaelClerx: Fixed them!

Issue #176, opened 2020-04-09
Mapping vs connection wording?


kerimoyle:
Should we be consistent about calling equivalent variables "mapped" (as they're formed form `map_variables` elements) and components "connected" (as they're formed from `connection` elements)?

MichaelClerx:
Hmmmm!

To me, the connection is between the variables, and the `<connection>` and `<map_variables>` elements are just a way we specify the details of these connections.

But going from the spec alone I see no reason you couldn't think of them as connections between components, within which lives mappings of variables...

I'd be careful to write from the second point of view, as I don't think that's what we intended in the rest of the doc, so it could lead to inconsistencies

Historically, in CellML 1 we had `<connection><map_components c1 c2 /><map_variables v1 v2 /></connection>` which perhaps made it clearer that we were specifying connections using a combination of `<map_x>` elements? In CellML 2 that's become `<connection c1 c2><map_variables v1 v2 /></connection>` which is shorter, but is not intended to imply there's some of of connection between components first, and some kind of further mapping between variables secondly. You can't have a connection without a map_variables, for example

kerimoyle:
In meeting on 12.4.20 decided to change "mapped/connected variables" to "equivalent variables".

PR #210, opened 2020-04-14
i176 Change connected/mapped variables into equivalent variables

kerimoyle:
Addresses #176 

Rendered version is here: https://cellml-specification-dev.readthedocs.io/en/i176_connected_to_equivalent_variables/reference/formal_and_informative/specB09.html


src/reference/sectionC_interpretation.inc
```diff
@@ -383,9 +383,9 @@ Interpretation of ``map_variables`` elements
 
 #.  CellML models MUST only contain :code:`map_variables` elements where the interface of variable *A* and the interface of variable *B* are applicable interfaces.
 
-#.  The :code:`variable` elements in a CellML model SHALL be treated as belonging to a single "connected variable set".
-    Each set of connected variables is the set of all :code:`variable` elements for which the corresponding nodes in the variable equivalence network form a connected subgraph.
-    Each set of connected variables represents one variable in the underlying mathematical model.
+#.  The :code:`variable` elements in a CellML model SHALL be treated as belonging to a single "equivalent variable set".
+    Each set of equivalent variables is the set of all :code:`variable` elements for which the corresponding nodes in the variable equivalence network form a equivalent subgraph.
```
nickerso, line 387: I think that final connected is actually talking about the connected nodes in the network rather than the connected/equivalent issue :)

Issue #207, opened 2020-04-14
Mixing of XML/info-item talk and conceptual talk


MichaelClerx:
Hi all!

## Problem

In recent work on chapter 3, we've started mixing talk of syntax and XML items with talk of conceptual items i.e. the thing a CellML model represents. 

It's visible most clearly when we talk about references (which are now apparently to XML elements, rather than to variables, components, etc.). A reference is said in many places to be to e.g. "a `units` element", rather than to the conceptual unit it represents. This makes all the language really weird and convoluted, and turns chapter 3 into a second chapter on syntax (which was what chapter 2 was supposed to be).

In sections like 3.9 and 3.10 we do it much better, and say things like

> For **a given variable**, the available interfaces SHALL be determined by the interface attribute information item **on the corresponding variable element** as follows.

## Solution

I think we can fix this quite easily, by 

- re-ordering chapter 3 a bit, so that we have the space to say things like "a model element defines a model", "a variable element defines a variable", before we get into e.g. encapsulation and imports
- re-writing parts of chapter 3, without changing its meaning

It's a bit of work, but it would ensure the spec says what we intend it to say. It wouldn't actually change its (intended) meaning (just its actual meaning).

## Alternative?

The alternative is to do this later, in some kind of CellML 2.0.1. This would be really annoying as this second spec wouldn't actually introduce or remove anything from the language, it would just describe it better.

MichaelClerx:
It would also help us get rid of the "units reference" and "component reference" and "variable reference" sections, which are really annoying, and of the "effect of a units attribute" section which should really just be part of "interpretation of variable"

MichaelClerx:
Old thoughts in #206

MichaelClerx:
It would look something like this

### Interpretation of models
A _model_ is a conceptual entity defining a system of equations.
A model is defined by a :code:`model` element.
A model has a _name_, determined by the :code:`name` attribute of its :code:`model` element.
A model contains zero or more _units_ [LINK].
A model contains zero or more _components_ [LINK].

### Interpretation of units
A _units definition_ is a conceptual entity representing physical units.
A units definition is defined by a :code:`units` element or an :code:`import units` element, or is one of the predefined units of table X.
A units definition has a _name_, determined by the :code:`name` attribute of its :code:`units` element or :code:`import units` element, or by the "Name" column in table X for predefined units.
A units definition has a _unit reduction_, determined by the [LINK]rules in section XYZ.

[...]

PR #211, opened 2020-04-15
i207 Reorder and rewrite chapter 3

kerimoyle:
Addresses #207 

Rendered version available here: https://cellml-specification-dev.readthedocs.io/en/i207_reorder_rewrite_chapter_3/reference/index_sectionC.html


kerimoyle:
Hey @MichaelClerx!  I've put in everything (I think) but in your order.  

- The best way to change things will be do change the `sectionC_interpretation.inc` file, as everything is taken from there.  

- `sectionC_interpretation.inc` is still in the same order as previously, but that's not important because all that's controlled from the `index_sectionC.rst` file.  

- I also haven't changed what would be rendered in the normative only PDF - since it's a subset of the informative html files, it's easier to do that one first, then delete to get to the other instead of vv.

hsorby:
Closing in favour of #356 targeting the `main` branch.

Issue #187, opened 2020-04-10
Interpretation of imports, again.


MichaelClerx:
> The imported infoset SHALL be considered such that items relevant to the imported infoset (**unit children of imported units elements** or **the encapsulated set of imported component elements**) are themselves available to that import element child.

This text doesn't mention units used by the imported components

MichaelClerx:
Maybe we can split this point, 3.1.2, into two points: one saying that units references are to be interpreted w.r.t. to parent model, and a second point saying that encapsulated components are also imported

MichaelClerx:
Also, I can kiiiiinda see now how we thought only the first point was enough, so maybe the 2nd (and possibly 3d) point should be sub-points of the first? or the first can just go?

MichaelClerx:
I still find this section hard to read!

I'm trying to summarise it for myself, just to see if I understand it (and if it makes sense). This is the best I've been able to do so far:

> 3.1.1 An imported `units` or `component` should be interpreted in the context of its parent model.
> 3.1.2 Unit references in imported `units` or `components` elements should be interpreted in terms of their parent model.
> 3.1.2 However, units referenced by imported elements, but not explicitly imported themselves, remain unavailable to the importing model.
> 3.1.2 If an imported component is part of an encapsulation hierarchy, then all its child components and any connections to child component variables remain intact. This rule applies recursively.
> 3.1.2 Connections to any sibling components of an imported component are broken.

I'm not thinking that:

- 3.1.1 can go, in favour of a more explicit statement of what that means (e.g. the stuff I have as 3.1.2 above)
- 3.1.1 might actually be invalid? because "interpreting" a component could well mean looking at the encapsulation hierarchy and seeing its connected to siblings, or has a parent, and these would then be imported too, just like child components (because why would they be different?)

MichaelClerx:
I'm not sure if we need a rule saying you can't reference imported child components, because you can never reference child components outside of the context of the parent component anyway.
However, without such a rule maybe I could import a component, and then create new connections between it and its child components?

kerimoyle:
> 3.1.1 might actually be invalid? because "interpreting" a component could well mean looking at the encapsulation hierarchy and seeing its connected to siblings, or has a parent, and these would then be imported too, just like child components (because why would they be different?)

[...]

PR #212, opened 2020-04-15
Reworded import interpretation.

MichaelClerx:
Addresses #187


src/reference/sectionC_interpretation.inc
```diff
@@ -8,12 +8,15 @@ Interpretation of ``import`` elements
...
 
-#. The imported infoset SHALL be considered such that items relevant to the imported infoset (:code:`unit` children of imported :code:`units` elements or the :ref:`encapsulated set<specC_encapsulation_set>` of imported :code:`component` elements) are themselves available to that :code:`import` element child.
-   It is noted, for the avoidance of doubt, that the importing infoset SHALL NOT have access to such implicitly imported items, but only to explicitly specified :code:`import` element children.
+#. A :ref:`units reference<specC_units_reference>` occurring within an imported component or within an imported units definition, SHALL be resolved with respect to the imported infoset.
+
+#. When determining the equivalent variable set of a variable in an imported component:
+
+   #. connections to variables in the encapsulated set of the imported component SHALL be considered, and this rule SHALL be applied recursively;
+
+   #. connections to variables in the sibling set of the imported component SHALL NOT be considered.
```
kerimoyle, line 18: maintained

src/reference/sectionC_interpretation.inc
```diff
@@ -8,12 +8,15 @@ Interpretation of ``import`` elements
...
-   This imported infoset instance SHALL provide the context in which the child of an import element (a :code:`units` or :code:`component` element) is defined and interpreted.
 
-#. The imported infoset SHALL be considered such that items relevant to the imported infoset (:code:`unit` children of imported :code:`units` elements or the :ref:`encapsulated set<specC_encapsulation_set>` of imported :code:`component` elements) are themselves available to that :code:`import` element child.
-   It is noted, for the avoidance of doubt, that the importing infoset SHALL NOT have access to such implicitly imported items, but only to explicitly specified :code:`import` element children.
+#. A :ref:`units reference<specC_units_reference>` occurring within an imported component or within an imported units definition, SHALL be resolved with respect to the imported infoset.
+
+#. When determining the equivalent variable set of a variable in an imported component:
+
```
kerimoyle, line 15: In addition to the rules specified in ??, when determining the ... etc

MichaelClerx:
Have made the suggested changes, and made the wording a bit more precise.

- I say "connections between components" instead of "connections between variables" because otherwise I have to say "connections between variables from components in encapsulated sets" etc. I think it's enough to say "connections between components"?

- Added a bit saying if you import a child but not its parent then those connections are cut too

- There's something I didn't consider, which is connections between siblings that are both explicitly imported (or between a child and parent that are explicitly imported). If these are also maintained we'll need to add a bit saying "unless both components are explicitly imported" or something to that effect.

MichaelClerx:
>     * There's something I didn't consider, which is connections between siblings that are both explicitly imported (or between a child and parent that are explicitly imported). If these are also maintained we'll need to add a bit saying "unless both components are explicitly imported" or something to that effect.

Answer seems to be no, these don't get imported

MichaelClerx:
So what if I import component A, which has connections to its child B
_and_ I import child B explicitly

Then does component A keep talking to the old component B, but my model gets a new copy of B (B*)?
Because I could do things like make B* into A's parent in my new model

kerimoyle:
> So what if I import component A, which has connections to its child B
> and I import child B explicitly

You'll get two copies of B: one inside A (to which you don't have access) and your own separate copy (to which you do have access).

MichaelClerx:
> So what if I import component A, which has connections to its child B
> _and_ I import child B explicitly
> 
> Then does component A keep talking to the old component B, but my model gets a new copy of B (B*)?
> Because I could do things like make B* into A's parent in my new model

Because in that case, I'd have to adjust this a bit:

> Each :code:`import` element present in a :ref:`CellML infoset<specA_cellml_infoset>` (the importing infoset) SHALL define a new and distinct instance of the CellML infoset which is referenced by the :code:`href` attribute (the imported infoset).

so that it isn't each `import` element that's a new infoset, it's each `import component` or `import units`

kerimoyle:
> so that it isn't each import element that's a new infoset, it's each import component or import units

As I understand it, yes.  Each time you import anything you get a brand new infoset-for-context and a brand new thing.

MichaelClerx:
Thanks!

Thoughts @agarny ? Is this how you implemented them in the past? Is this how they are in libcellml @nickerso and @hsorby ?

MichaelClerx:
I've updated the wording to match the views expressed above!

I was tempted to add a "for the avoidance of doubt" saying that if you import A it keeps its links to child B, but then if you import B that's a *new* B, without links to your A. But I guess that's implied and so for the informative section!

src/reference/sectionC_interpretation.inc
```diff
@@ -7,13 +7,19 @@
...
 
-#. The imported infoset SHALL be considered such that items relevant to the imported infoset (:code:`unit` children of imported :code:`units` elements or the :ref:`encapsulated set<specC_encapsulation_set>` of imported :code:`component` elements) are themselves available to that :code:`import` element child.
-   It is noted, for the avoidance of doubt, that the importing infoset SHALL NOT have access to such implicitly imported items, but only to explicitly specified :code:`import` element children.
+#. A :ref:`units reference<specC_units_reference>` occurring within an imported component or within an imported units definition, SHALL be resolved with respect to the imported infoset.
 
-See :ref:`Units reference<specC_units_reference>` and :ref:`Component reference<specC_component_reference>` for the specifics of importing units and components.
+#. When determining the equivalent variable set of a variable in an imported component:
+
+   #. connections made in the importing infoset shall be handled as usual (see :ref:`Interpretation of ``map_variables`` <specC_interpretation_of_map_variables>`)
```
kerimoyle, line 16: shall be handled as described in Interpretation of blah blah

src/reference/sectionC_interpretation.inc
```diff
@@ -7,13 +7,19 @@
...
+
+   #. connections made in the imported infoset shall be handled as follows:
```
kerimoyle, line 18: Instead of saying "connections **made**" can we maybe say something that implies they already exist?  that we're not **making** them by the import?

src/reference/sectionC_interpretation.inc
```diff
@@ -7,13 +7,19 @@
...
+
+      #. connections to components in the encapsulated set of the imported component SHALL be maintained, and this rule SHALL be applied recursively;
+
+      #. connections to components in the sibling set, or to the encapsulation parent of the imported component SHALL NOT be maintained.
```
kerimoyle, line 22: Yep, that's pretty clear :)

src/reference/sectionC_interpretation.inc
MichaelClerx: Changed to 'defined'

src/reference/sectionC_interpretation.inc
MichaelClerx: Done!

src/reference/sectionC_interpretation.inc
MichaelClerx: Thanks!

src/reference/sectionC_interpretation.inc
```diff
@@ -7,13 +7,19 @@
...
 
-#. The imported infoset SHALL be considered such that items relevant to the imported infoset (:code:`unit` children of imported :code:`units` elements or the :ref:`encapsulated set<specC_encapsulation_set>` of imported :code:`component` elements) are themselves available to that :code:`import` element child.
-   It is noted, for the avoidance of doubt, that the importing infoset SHALL NOT have access to such implicitly imported items, but only to explicitly specified :code:`import` element children.
+#. A :ref:`units reference<specC_units_reference>` occurring within an imported component or within an imported units definition, SHALL be resolved with respect to the imported infoset.
 
-See :ref:`Units reference<specC_units_reference>` and :ref:`Component reference<specC_component_reference>` for the specifics of importing units and components.
+#. When determining the equivalent variable set of a variable in an imported component:
+
+   #. connections defined in the importing infoset shall be handled as described in :ref:`Interpretation of ``map_variables`` <specC_interpretation_of_map_variables>`.
```
kerimoyle, line 16: Does this formatting of the link build correctly?  I didn't think it would ... might have to change to:
```
:ref:`Interpretation of map_variables<specC_interpretation_of_map_variables>`
```
(no space before <, and no formatting allowed between the backticks)?

agarny: Correct, the formatting of the link doesn't build correctly. What works is (done in commit e574934):
```
:ref:`Interpretation of map_variables<specC_interpretation_of_map_variables>`
```
(FWIW, whether there is a space of or not before `<` doesn't make any difference.)

Still, ideally, "map_variables" would be rendered using code style, i.e. "`map_variables`". I know this can be done with when we have an external link (see https://opencor.ws/developer/thirdPartyLibraries.html), but can't remember / don't know how to do it for a cross reference.

src/reference/sectionC_interpretation.inc
```diff
@@ -7,13 +7,19 @@
...
-   This imported infoset instance SHALL provide the context in which the child of an import element (a :code:`units` or :code:`component` element) is defined and interpreted.
+#. Each :code:`import units` or :code:`import component` element present in a :ref:`CellML infoset<specA_cellml_infoset>` (the importing infoset) SHALL define a new and distinct instance of the CellML infoset (the imported infoset) that is specified by the parent :code:`import` element's :code:`href` attribute.
 
-#. The imported infoset SHALL be considered such that items relevant to the imported infoset (:code:`unit` children of imported :code:`units` elements or the :ref:`encapsulated set<specC_encapsulation_set>` of imported :code:`component` elements) are themselves available to that :code:`import` element child.
-   It is noted, for the avoidance of doubt, that the importing infoset SHALL NOT have access to such implicitly imported items, but only to explicitly specified :code:`import` element children.
+#. A :ref:`units reference<specC_units_reference>` occurring within an imported component or within an imported units definition, SHALL be resolved with respect to the imported infoset.
```
hsorby, line 12: > occurring within an imported component

occurring within an imported components encapsulation set.

MichaelClerx: occurring within an imported element ?

PR #215, opened 2020-04-15
Slight change to reset wording. 

MichaelClerx:
Taking Keri's comments from #191 on board

### DIFF
--- a/src/reference/sectionC_interpretation.inc
+++ b/src/reference/sectionC_interpretation.inc
@@ -410,12 +410,14 @@ Interpretation of ``map_variables`` elements
 Interpretation of ``reset`` elements
 ------------------------------------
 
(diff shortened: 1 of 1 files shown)


src/reference/sectionC_interpretation.inc
```diff
@@ -410,12 +410,14 @@ Interpretation of ``map_variables`` elements
 Interpretation of ``reset`` elements
 ------------------------------------
 
-#. Each :code:`reset` element describes a change to be applied to the variable referenced by the :code:`variable` attribute when specified conditions are met during the simulation of the model.
+#. For the purposes of this section, we define the "reset variable" to be the variable referenced by a :code:`reset` element's :code:`variable` attribute, and the "test variable" to be the variable referenced by its :code:`test_variable` attribute.
```
kerimoyle, line 413: Is it just in this section or throughout the spec?

MichaelClerx: Just in this section. They're not discussed anywhere else

Issue #197, opened 2020-04-14
Various minor issues with 2.0-rc1


agarny:
I have several minor issues that I didn't think deserved an issue of their own:
- General:
  - [x] #214 Check all the links (a lot of them are broken). 
  - [x] Is there any rationale for our use of italics? Yes, see https://github.com/cellml/cellml-specification/issues/197#issuecomment-614380841
  - [x] #218 Sometimes "CellML namespace" is a link, sometimes it is not. The first occurrence of "CellML namespace" in a section should be a link. 
  - [x] #220 Be consistent in our use of numbered lists, i.e. use ";" at the end of an item rather than ".", except for the last item. May also decide on whether to use a final "and"/"or" (e.g., see 1.2.2.1 vs. 1.2.5.1 vs. 1.3.1 vs. 2.1.2). 
- Contents:
  - [ ] #221 Move 2.12 (`math` element) to 2.9, i.e. before `reset` element? It's just that we can perfectly have a model without resets, so the `reset` & Co. elements should be considered as "add-ons" and I would thefeore expect them to come after the `math` element. 
- Preample:
  - [x]  #227 "A PDF [version] of this [document] is available [at]..."
  - [x] (discussion below) ~Add a full stop at the end of "Contact: editors@cellml.org".~
- 1.1 Terminology:
  - [x] #222 No need to mention the "REQUIRED", "SHOULD", "SHOULD NOT", "RECOMMENDED" and "OPTIONAL" keywords. They are never used in the spec. 
  - [x] #222 CellML infoset: "In this specification[,] such infosets are assumed to be CellML 2.0 infosets." (missing comma)
  - [x] ~CellML model: Use "top-most" or then "toplevel". We use "top-level" somewhere else, so probably "top-most" then.~
  - [x] #222 Whitespace character: list the Unicode characters in order, i.e. U+0009, U+000A, U+000D or U+0020.
- 1.2 CellML information sets:
  - [x] (discussion below) 1.2.3.1: replace "... making any or none of..." with "... making none or any of..."?
  - [x] #223 Does 1.2.4.1 make sense (Character information items)?
  - [x] #220 1.2.5.1: add ";" at the end of the first namespace  and "." at the end of the second namespace.
  - [x] #201 1.2.5.1: are we supporting other namespaces besides the CellML and MathML ones? A quick search tells me that we don't, so no need for ", unless explicitly indicated otherwise".
[...]

PR #222, opened 2020-04-15
Updates to terminology section wording as suggested in #197.

MichaelClerx:
- No need to mention the "REQUIRED", "SHOULD", "SHOULD NOT", "RECOMMENDED" and "OPTIONAL" keywords. They are never used in the spec.
  - CellML infoset: "In this specification[,] such infosets are assumed to be CellML 2.0 infosets." (missing comma)
  - Whitespace character: list the Unicode characters in order, i.e. U+0009, U+000A, U+000D or U+0020.

(w.r.t. to the comma, we do it elsewhere in the same chapter, so doing it here too)

### DIFF
--- a/src/reference/sectionA_definitions.inc
+++ b/src/reference/sectionA_definitions.inc
@@ -7,7 +7,7 @@
 Terminology
 -----------
 
-The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in :rfc2119:`RFC 2119`.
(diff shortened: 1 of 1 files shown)


MichaelClerx:
> (I shall do my best to silence my grumblings about comma usage ;) )

:D

I'd quite like to say "whoever writes a section decides what they do in there stylistically", but if there's two sentences inside a section I'd like them to be consistent, or if there's one sentence that gets repeated in every section. Otherwise lets just go with whatever someone writes down first :D

agarny:
Hmm... we need consistency throughout the shole spec, not only within a section. Anything else will make the spec look amateurish to me.

MichaelClerx:
I'm sure that if you look at the comments in the linux kernel they won't all be stylistically the same

agarny:
The Linux kernel is not a specification document...

PR #223, opened 2020-04-15
Change to whitespace-in-elements wording

MichaelClerx:
A char info item is a single char. See #197.

### DIFF
--- a/src/reference/sectionA_definitions.inc
+++ b/src/reference/sectionA_definitions.inc
@@ -142,7 +142,7 @@ Non-specific information items
 
    #. A character information item.
 
-2. An element information item in the :ref:`CellML namespace<specA_cellml_namespace>` MUST NOT contain any character information items, except for character information items which consist entirely of whitespace characters.
+2. An element information item in the :ref:`CellML namespace<specA_cellml_namespace>` MUST NOT contain any character information items, except for whitespace characters.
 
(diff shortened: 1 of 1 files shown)


src/reference/sectionA_definitions.inc
```diff
@@ -142,7 +142,7 @@ Non-specific information items
 
    #. A character information item.
 
-2. An element information item in the :ref:`CellML namespace<specA_cellml_namespace>` MUST NOT contain any character information items, except for character information items which consist entirely of whitespace characters.
+2. An element information item in the :ref:`CellML namespace<specA_cellml_namespace>` MUST NOT contain any character information items, except for whitespace characters.
```
kerimoyle, line 145: Do we need the "information items" bit at all?

2. An element information item in the :ref:`CellML namespace<specA_cellml_namespace>` MUST NOT contain any characters other than whitespace characters.

MichaelClerx: A "character information item" is an XML thing. It means a character in a file, but also some other more vague things than that. So yes, is my opinion. We're talking in XML information item terms in this section

kerimoyle: Okeydokes.

kerimoyle:
ping @agarny

Issue #204, opened 2020-04-14
"References to component elements" discusses import mechanism


MichaelClerx:
If anywhere, this should be in the interpretation of imports elements section

MichaelClerx:
Will be addressed by #206 , but will try to do a back-up fix too

Issue #189, opened 2020-04-10
The component reference and variable reference sections are redundant?


MichaelClerx:
We say what the rules are when we use component references, and then we tack on a (see component reference) bit where we re-explain but worse?

MichaelClerx:
Also pretty sure this isn't right:

> The term “component reference” refers to **the value of a component_ref attribute used in an import item for component items**, to component_ref elements which occur as children of encapsulation items, and to the value of component_1 and component_2 attributes used in connection items.
>
> 1. A component reference SHALL be the name of a component, and SHALL be interpreted based on the context within the CellML model in which it occurs.

What is 1 saying here? When you import a component that's obviously a reference to the model you're importing from, and you need to interpret the component in that model's context. So that's not it.

So then what is "interpreting a component reference" in this context? And why do we need to specify you need to interpret part of a model in the context of that model to begin with?

MichaelClerx:
Trying to summarise 3.4 and after 30 minutes I have literally nothing. I don't see anything that this section adds to the stuff already said in chapter 2!

MichaelClerx:
Actually 3.4.4 seems new, but why are we discussing the details of imports in this section? I find it really strange!

MichaelClerx:
3.5 similarly immediatly breaks up into context-specific rules that are already in chapter 2 ?

MichaelClerx:
Unit references are tricky, and imports are tricky, but the rest is just string matching?
Don't really see the point of complicating simple references within a model by trying to describe it at the same time as the import stuff

MichaelClerx:
I'd also say these "component reference" and "variable reference" sections aren't really about interpretation, they're just giving basic syntax

MichaelClerx:
Will be addressed by #206 , but will try to do a back-up fix too

Issue #188, opened 2020-04-10
Units reference wording


MichaelClerx:
> 1. A units reference SHALL be a CellML identifier** and SHALL be interpreted dependent on the context of the CellML model in which it occurs, according to the units referencing rules defined later in this section**.

What is that bold bit all about? *Everything* in a cellml model "shall be interpreted dependent on the context of the model in which it occurs", and everything in a model shall be interpreted according to the rules in the spec. Why on earth are we stating that?

MichaelClerx:
Does this really need to be so complicated?

> 1. Where, within the same infoset, a units element has a name attribute identical to the units reference, the units reference SHALL refer to that units element.
> 2. Where there is an import units element in the CellML infoset, such that the import units
element has a name attribute identical to the units reference, then the units reference SHALL be treated with respect to referencing rules as if the units reference appeared in the imported infoset, and referring to the name specified in the units_ref attribute of the import units element.
> 3. Where the units reference is equal to the value in the “Name” column of the Built-in units table, then the units reference SHALL be a reference to the built-in units corresponding to that row of the table.

All we're saying is a units reference must be to (3) a built-in unit, (2) the name of a "units" in the same model, or (3) the name of an "import units" in the same model ?

MichaelClerx:
The current text doesn't seem to say that these 3 options are exhaustive, it just describes 3 scenarios without stating they're the only ones allowed

MichaelClerx:
Also still don't get why we have a "symbol" and a "common definition" column. They are never mentioned in the spec, you don't need them to interpret the spec, they're not part of CellML. We don't have a nice info block with the history of variables either, or some background on MathML.

kerimoyle:
They're both a consequence of the radians debate :). In order to say that radians are dimensionless but really length/length, we added the common definitions column.  In order to make that column actually readable, we added the symbol column ... I think that while it's not strictly necessary it would help people understand the ideas behind unit reduction, especially when the unit reduction is in a form they're not used to seeing.  I don't see the harm?

> Everything in a cellml model "shall be interpreted dependent on the context of the model in which it occurs", and everything in a model shall be interpreted according to the rules in the spec. Why on earth are we stating that?

[...]

PR #230, opened 2020-04-15
Updates to unit, component, and variable reference interpretation

MichaelClerx:
See #188, #189, and #204

This is an attempt to clean up the "reference to" sections in chapter 3, without the "proper" fixes of #207. As a stop-gap in case we don't have time to work everything out.

### General notes:

- This branches off of #212, because it needs the updated import wording
- Changed 'valid CellML identifier' to 'CellML identifier' because we don't define what an invalid CellML identifier could be

### Units reference notes:

- _"The :code:`units` element inclusion digraph SHALL ... that is equal to the :code:`name` attribute value of :code:`units` element *B*."_
  - This used to involve a "units reference", but that reference should really point to a unit concept, not a `units` element, but that meant it could also involve built-ins and imported units
  - This graph does _not_ include built-in units, but this is OK, as these cannot be used to create cycles. Similarly, `import units` elements are not included in the graph, but again these cannot be used to create cycles (because imports cannot be cyclical, imported units can never refer to units from the importing model).
- Removed the text in C saying where a units reference occurs
  - Firstly, because that's cyclical: we say "X must be a units ref", then we shouldn't define a units ref as "the thing X should be".
  - Secondly, the list was wrong. A units reference occurs as the `units` attribute of a `variable`, the `units` attribute of a `unit`, or the `units` attribute of a `cn`. The value of a :code:`units_ref` attribute is NOT a units reference. And never has been in section B. The difference is that a units reference is to a built-in, local, or imported unit, while a `units_ref` is to either a unit or an imported unit _in an imported infoset_. That distinction is important when writing definitions of stuff.
- The current definition says a units ref is to 1 of 3 things, and if it's thing 3 (an import units), then the `units_ref` of that import units should be treated as a units ref within the imported infoset.
  - I hope it's clear that this is a recursive rule, but have emphasised so anyway
  - This allows importing a built-in unit, as long as you rename it. To me that's the same as introducing a new local unit that's equal to a built-in but renamed, so don't see why we'd disallow it if that means adding extra text.
- We don't need to say something like "a `units` element that occurs as a child of a `model` element" because that's how a units element is defined. The other one is called an `import units` element.

### Component reference notes
- Again, the "component_ref" on an `import component` is not a component reference.
- Again, we distinguish between the two `<component>` tags by calling them a `component` element and an `import component` element.

### Variable reference notes
- Same again.


src/reference/sectionC_interpretation.inc
```diff
@@ -7,33 +7,41 @@
 Interpretation of ``import`` elements
 -------------------------------------
 
-#. Each :code:`import` element present in a :ref:`CellML infoset<specA_cellml_infoset>` (the importing infoset) SHALL define a new and distinct instance of the CellML infoset which is referenced by the :code:`href` attribute (the imported infoset).
-   This imported infoset instance SHALL provide the context in which the child of an import element (a :code:`units` or :code:`component` element) is defined and interpreted.
+#. Each :code:`import units` or :code:`import component` element present in a :ref:`CellML infoset<specA_cellml_infoset>` (the importing infoset) SHALL define a new and distinct instance of the CellML infoset (the imported infoset) that is specified by the parent :code:`import` element's :code:`href` attribute.
```
MichaelClerx, line 10: To comment on this bit see #212

src/reference/sectionC_interpretation.inc
```diff
@@ -7,33 +7,41 @@
...
 
-#. The imported infoset SHALL be considered such that items relevant to the imported infoset (:code:`unit` children of imported :code:`units` elements or the :ref:`encapsulated set<specC_encapsulation_set>` of imported :code:`component` elements) are themselves available to that :code:`import` element child.
-   It is noted, for the avoidance of doubt, that the importing infoset SHALL NOT have access to such implicitly imported items, but only to explicitly specified :code:`import` element children.
+#. A :ref:`units reference<specC_units_reference>` occurring within an imported component or within an imported units definition, SHALL be resolved with respect to the imported infoset.
```
MichaelClerx, line 12: To comment on this bit see #212

src/reference/sectionC_interpretation.inc
```diff
@@ -7,33 +7,41 @@
...
 
-See :ref:`Units reference<specC_units_reference>` and :ref:`Component reference<specC_component_reference>` for the specifics of importing units and components.
+#. When determining the equivalent variable set of a variable in an imported component:
```
MichaelClerx, line 14: To comment on this bit see #212

src/reference/sectionC_interpretation.inc
```diff
@@ -7,33 +7,41 @@
...
+
+   #. connections defined in the importing infoset shall be handled as described in :ref:`Interpretation of ``map_variables`` <specC_interpretation_of_map_variables>`.
+
+   #. connections defined in the imported infoset shall be handled as follows:
+
+      #. connections to components in the encapsulated set of the imported component SHALL be maintained, and this rule SHALL be applied recursively;
+
+      #. connections to components in the sibling set, or to the encapsulation parent of the imported component SHALL NOT be maintained.
```
MichaelClerx, line 22: End of stuff from #212

src/reference/sectionB_elements.inc
```diff
@@ -115,16 +115,17 @@ An :code:`import component` element information item (referred to in this specif
 
 .. container:: issue-import-component-name
 
-   1. Every :code:`import component` element MUST contain a :code:`name` attribute. The value of the :code:`name` attribute MUST be a :ref:`valid CellML identifier<specA_cellml_identifier>`.
+   1. Every :code:`import component` element MUST contain a :code:`name` attribute.
+      The value of the :code:`name` attribute MUST be a :ref:`CellML identifier<specA_cellml_identifier>`.
 
-      The value of the :code:`name` attribute MUST NOT be identical to the :code:`name` attribute of any other :code:`component` or :code:`import component` element in the :ref:`CellML infoset<specA_cellml_infoset>`.
+      The value of the :code:`name` attribute MUST NOT be identical to the :code:`name` attribute of any :code:`component` or :code:`import component` element in the :ref:`CellML infoset<specA_cellml_infoset>`.
```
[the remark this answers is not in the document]
MichaelClerx: Fair enough, I'll stick it back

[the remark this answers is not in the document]
MichaelClerx: And double-check the others

MichaelClerx:
Approved in meeting after above change is made

Issue #214, opened 2020-04-15
Fix all external links


MichaelClerx:
- [x] Stuff is getting appended where it shouldn't be
- [x] Two links to infoset (in sectionA_definitions) should be moved to config file
- [x] Links to informative spec in preamble should be in config file
- [x] Decision re dated or canonical URL for references and display

kerimoyle:
We have different addresses for the infoset links:

https://www.w3.org/TR/2004/REC-xml-infoset-20040204/
https://www.w3.org/TR/xml-infoset/

... which one is better?

kerimoyle:
Ditto namespaces: 
https://www.w3.org/TR/2006/REC-xml-names11-20060816/ 
https://www.w3.org/TR/xml-names11/ 

And MathML2: 
http://www.w3.org/TR/2003/REC-MathML2-20031021/
https://www.w3.org/TR/MathML2/

kerimoyle:
Xlink: 
https://www.w3.org/TR/xlink11/
http://www.w3.org/TR/2001/REC-xlink-20010627

MichaelClerx:
They seem to start with

> This version:
>     http://www.w3.org/TR/2004/REC-xml-infoset-20040204 
> Latest version:
>     http://www.w3.org/TR/xml-infoset 

but not sure that helps?

Maybe the w3c is a slower glacier than even CellML, so the latest version will be "the same but worded better"

But maybe we should play safe and just go for the one that we can see right now?

kerimoyle:
I should have asked the question better! There are three types of references:

- Those hardcoded in the text, intended for full display, eg in namespaces: do we want to **display** in the text the latest or the dated URL?
- Those in the reference section at the end: do we want to **display** in the references the latest or the dated URL (answered in part in #123)
- Those which are the underlying clickable links, where the URL is not displayed: do we want to **send** people to the latest or the dated URL?

MichaelClerx:
I'd say the same for all three, to avoid confusion.

And leaning towards dated for all three.

Also just leaning in general. Goodnight!

agarny:
Agreed that the URL should be the same in all three cases.

Re dated vs non-dated URLs, I like short URLs, but... my preference would still be for dated URLs, so as to "protect" ourselves against future potential changes that might break things for us.

kerimoyle:
[...]

PR #231, opened 2020-04-16
i214 fix external links

kerimoyle:
Addresses #214 

Rendered version here: https://cellml-specification-dev.readthedocs.io/en/i214_external_links/

Still required: 
- [x] Real address for PDF?
- [x] Real address for informative spec?
- [x] Decision as to which versions of URLs to display/link to (dated or not)


kerimoyle:
I'm not superconfident that I'll build the PDF correctly ... please can someone check that the preamble links appear as you'd expect? (They may not be the correct addresses at the minute, but they should at least be rendered correctly).
Other than that, the references through the text should be correct now, as should the Reference section at the end.

agarny:
> I'm not superconfident that I'll build the PDF correctly ... please can someone check that the preamble links appear as you'd expect? (They may not be the correct addresses at the minute, but they should at least be rendered correctly).
> Other than that, the references through the text should be correct now, as should the Reference section at the end.

Here is the [PDF file](https://github.com/cellml/cellml-specification/files/4484842/cellml2_specification.pdf) I got after disabling `hoverxref`. Here is the corresponding [log file](https://github.com/cellml/cellml-specification/files/4484843/log.txt).

kerimoyle:
Thanks @agarny ... I've fixed that link.  Please could you try it with the normative spec only?  That's the one which the normative prelim bit will be in that I can't see?

agarny:
Oops, my bad. Here goes: [PDF](https://github.com/cellml/cellml-specification/files/4484966/cellml2_normative_specification.pdf) and [log](https://github.com/cellml/cellml-specification/files/4484967/log.txt).

kerimoyle:
Gah, I still don't understand these link errors in the PDF.  They're fine in the html :( all I've changed in fefd1b2 is the display.  Any ideas ... ?

agarny:
> Gah, I still don't understand these link errors in the PDF. They're fine in the html :( all I've changed in [fefd1b2](https://github.com/cellml/cellml-specification/commit/fefd1b29ca9cc307a0e29a98893a023db6160be8) is the display. Any ideas ... ?

That's because you updated your branch and I didn't pull things... Argh! New [PDF](https://github.com/cellml/cellml-specification/files/4485032/cellml2_normative_specification.pdf) and new (better looking) [log](https://github.com/cellml/cellml-specification/files/4485033/log.txt).

PR #244, opened 2020-04-16
i243 remove hash from rfc link

kerimoyle:
Addresses #243 

Rendered version here: https://cellml-specification-dev.readthedocs.io/en/i243_links_hash/reference/index_sectionD.html


kerimoyle:
https://cellml-specification-dev.readthedocs.io/en/i243_links_hash/reference/index_sectionD.html

PR #251, opened 2020-04-16
Merge cellml-2-drafting into master

### LINKED ISSUE #226: Replace contents of `master` with that of `cellml-2-drafting
I don't know who's looking at this repo anyway

MichaelClerx:
Assuming this branched off it at some point :D 

Closes #226 

Note that the overview below only shows the 250 latest commits (because, github helpfully told me, this PR is "big")


kerimoyle:
I'm not sure it did ... I created this from somewhere else, it wasn't related to the existing spec at all  ... not sure what difference that makes now though!

MichaelClerx:
I checked in the network graph, looks like it should indeed be right :D

PR #267, opened 2020-04-16
Various small fixes

MichaelClerx:
Fixes for various new tickets about details in the normative spec.

Most by keri, some by me.

### DIFF
--- a/src/reference/sectionC_interpretation.inc
+++ b/src/reference/sectionC_interpretation.inc
@@ -358,8 +358,6 @@ Interpretation of ``encapsulation`` elements
 
 .. marker_interpretation_of_encapsulation_2
 
-7. For the avoidance of doubt, there MUST NOT be a :code:`connection` element such that the component referenced by the :code:`component_1` attribute is in the hidden set of the component referenced by the :code:`component_2` attribute, nor vice versa.
-
 .. marker_interpretation_of_encapsulation_end
 .. marker_interpretation_of_map_variables_start
 
--- a/src/reference/sectionC_interpretation.inc
+++ b/src/reference/sectionC_interpretation.inc
@@ -391,7 +391,7 @@ Interpretation of ``map_variables`` elements
 
 .. marker_interpretation_of_map_variables_3
 
-6.  For a given variable, the available interfaces SHALL be determined by the :code:`interface` attribute on the corresponding :code:`variable` element as follows:
+6.  For a given variable, the "available interfaces" SHALL be determined by the :code:`interface` attribute on the corresponding :code:`variable` element as follows:
 
     #. A value of :code:`public` specifies that the variable has a public interface.
 
@@ -403,7 +403,7 @@ Interpretation of ``map_variables`` elements
 
     #. If the :code:`interface` attribute is absent, then the variable has no interface.
 
-7.  The applicable interfaces for variables *A* and *B* in components *AA* and *BB* respectively SHALL be defined as follows:
+7.  The "applicable interfaces" for variables *A* and *B* in components *AA* and *BB* respectively SHALL be defined as follows:
 
     #. When component *AA* is in the :ref:`sibling set<specC_sibling_set>` of component *BB* (and vice versa), the applicable interface for both variables *A* and *B* SHALL be the public interface.
 
@@ -411,9 +411,9 @@ Interpretation of ``map_variables`` elements
 
     #. When component *BB* is in the :ref:`encapsulated set<specC_encapsulation_set>` of component *AA* the applicable interface for variable *B* SHALL be the public interface, and the applicable interface for variable *A* SHALL be the private interface.
 
-    #. It is noted, for the avoidance of doubt, that any other encapsulation relationship between components *AA* and *BB* will not permit connections, meaning that no interface is available between the variables *A* and *B*.
+    It is noted, for the avoidance of doubt, that if components *AA* and *BB* are in each other's :ref:`hidden set<specC_hidden_set>`, there are no applicable interfaces for the variables *A* and *B*.
 
-8.  CellML models MUST only contain :code:`map_variables` elements where the interface of variable *A* and the interface of variable *B* are applicable interfaces.
+8.  CellML models MUST only contain :code:`map_variables` elements where the applicable interfaces for both variables are available.
 
 .. marker_interpretation_of_map_variables_4
 
--- a/src/reference/sectionC_interpretation.inc
+++ b/src/reference/sectionC_interpretation.inc
@@ -154,7 +154,7 @@ Interpretation of ``units`` elements
       If a :code:`unit` element has no :code:`multiplier` attribute, the multiplier term SHALL have value 1.0.
       Otherwise, the value of the :code:`multiplier` attribute MUST be a :ref:`real number string<specA_real_number>`, and the value of the multiplier term SHALL be the numerical value of that string.
 
-   4. The relationship between the product, :math:`P`, of numerical values given in each and every child :code:`unit` element units, to a numerical value, :math:`x`, with units given by the encompassing :code:`units` element, SHALL be
+   4. The relationship between the product, :math:`P`, of numerical values given in each and every child :code:`unit` element's :code:`units` attribute, to a numerical value, :math:`x`, with units given by the encompassing :code:`units` element, SHALL be
 
       .. math::
 
--- a/src/reference/sectionC_interpretation.inc
+++ b/src/reference/sectionC_interpretation.inc
@@ -286,11 +286,11 @@ Interpretation of ``initial_value`` attributes
 
 #. The :code:`initial_value` attribute of a :code:`variable` element MUST either be a :ref:`real number string<specA_real_number>`, or a :ref:`variable reference<specC_variable_reference>`.
 
-#. The conditions when initial values hold are (by design) not defined in a :ref:`CellML model<specA_cellml_model>`.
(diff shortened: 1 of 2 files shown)


src/reference/sectionC_interpretation.inc
```diff
@@ -167,33 +167,38 @@ Interpretation of ``units`` elements
...
+   The set of tuples SHALL be determined as follows:
 
    1. If the :code:`units` element has no :code:`unit` child elements, then the set of tuples SHALL have a single member, which SHALL consist of the name of the :code:`units` element and the exponent 1.0.
 
    2. If the :code:`units` element has one or more :code:`unit` child elements, then the set of tuples SHALL consist of the entire collection of tuples given by all :code:`unit` child elements.
+
       Tuples for each :code:`unit` child element SHALL be determined as follows:
 
-      1. Where the :ref:`units reference<specC_units_reference>` of the :code:`unit` child element is to a single unit which is an irreducible unit, then the set of tuples SHALL have a single member, which SHALL consist of the name of the irreducible unit being referenced and the exponent 1.0.
+      1. If the :ref:`units reference<specC_units_reference>` of the :code:`unit` child element is to a single unit which is an irreducible unit, then the set of tuples SHALL have a single member, which SHALL consist of the name of the irreducible unit being referenced and the exponent 1.0.
```
kerimoyle, line 188: Gah, this is why I so dislike "units" and "unit" terms ... how can we make it clear that the bold below is talking about physical units (singular) rather than the unit item? 

If the :ref:`units reference<specC_units_reference>` of the :code:`unit` child element is **to a single unit which is an irreducible unit**, then the set of tuples SHALL have a single member, which SHALL consist of the name of the irreducible unit being referenced and the exponent 1.0.

src/reference/sectionC_interpretation.inc
```diff
@@ -167,33 +167,38 @@ Interpretation of ``units`` elements
...
 
       4. The exponents of each tuple in the set for the current :code:`unit` element, as derived by following rules :hardcodedref:`3.3.3.2.1, 3.3.3.2.2 or 3.3.3.2.3` above, SHALL be multiplied by the exponent term of the current, referencing, :code:`unit` element.
 
-   3. Tuples which have the name element of "dimensionless" SHALL be removed from the set of tuples.
+   3. Tuples which have the unit name of "dimensionless" SHALL be removed from the set of tuples.
       Note that this can result in the set of tuples being empty.
 
-   4. Where the set of tuples consists of tuples which have the same name element, those tuples SHALL be combined into a single tuple with that name element and an exponent being the sum of the combined tuples’ exponents.
-      If the resulting tuple’s exponent term is zero, the tuple SHALL be removed from the set of tuples.
+   4. If the set of tuples contains tuples which have the same unit name, then those tuples SHALL be combined into a single tuple with that unit name and an exponent being the sum of the combined tuples’ exponents.
```
kerimoyle, line 200: Do we need "sum" as well as "combined"?  
Suggest: "... an exponent being the sum of those tuples' exponents."

src/reference/sectionC_interpretation.inc
```diff
@@ -281,9 +286,9 @@ Interpretation of ``initial_value`` attributes
 
 #. The :code:`initial_value` attribute of a :code:`variable` element MUST either be a :ref:`real number string<specA_real_number>`, or a :ref:`variable reference<specC_variable_reference>`.
 
-#. The conditions when initial values hold are (by design) not defined in a :ref:`CellML model<specA_cellml_model>` document.
+#. The conditions when initial values hold are (by design) not defined in a :ref:`CellML model<specA_cellml_model>`.
```
kerimoyle, line 289: Suggest: The conditions under which initial values hold ...
Dunno why, something about this bit sounds odd.

kerimoyle: Eh, but then we use "when" below too ... maybe just ignore me!

MichaelClerx: I suppose "when x = 4" is still a "when"?

kerimoyle: Yes, when x = 4 is fine, but "conditions when" sounds odd.  Like I said, I don't know why!

src/reference/sectionC_interpretation.inc
kerimoyle: Yeah, I know.  If everyone reads it with as much attention to font as we do, we'll all be ok ... ;)

agarny: FWIW, I am ok with the current text.

src/reference/sectionC_interpretation.inc
agarny: Agreed with @kerimoyle's suggestion.

src/reference/sectionC_interpretation.inc
MichaelClerx: I don't want to introduce new terminology either

MichaelClerx: I'm gonna hit resolve on these are they're not about the changes in the PR!

src/reference/sectionC_interpretation.inc
MichaelClerx: I think having combined in there is more clear

src/reference/sectionC_interpretation.inc
agarny: I prefer @kerimoyle's suggestion, whatever the implications on the rest of the text.

MichaelClerx: Have committed a change!

src/reference/sectionC_interpretation.inc
MichaelClerx: There's a full set of tuples, and this bit is about the subset of "combined" tuples with the same name, so worth keeping that in there to distinguish?
As in, it's not the sum of _all_ tuples exponents, just of the ones with the same name.

src/reference/sectionC_interpretation.inc
```diff
@@ -281,9 +286,9 @@ Interpretation of ``initial_value`` attributes
...
 
-#. Where the :code:`initial_value` attribute has a :ref:`real number value<specA_real_number>`, it SHALL be interpreted as a statement that the variable on which the attribute appears is equal to that real number value, under the conditions when the initial value holds.
+#. Where the :code:`initial_value` attribute is a :ref:`real number string<specA_real_number>`, it SHALL be interpreted as a statement that the variable on which the attribute appears is equal to that real number value, under the conditions when the initial value holds.
```
agarny: We previously replaced "Where" with "If" at the beginning of a clause. Why don't we do it here (and below, and maybe elsewhere too?)?

Similarly, there are been corrections that read "..., [then] it SHALL..." So, why no "then" here?

IOW, some recent changes have the risk of making our spec inconsistent. So, need to check that.

src/reference/sectionC_interpretation.inc
kerimoyle: Yep, hence the those: "sum of those tuples' exponents"

src/reference/sectionC_interpretation.inc
kerimoyle: I think it's going to be easiest to merge this and other PRs in first then check for consistency?

MichaelClerx: Anyway. Here a "where" is good. We've just had a sentence talking about _all_ intial values, so now we say "where" to indicate a specific one

src/reference/sectionC_interpretation.inc
```diff
@@ -400,23 +403,23 @@ Interpretation of ``map_variables`` elements
...
 
     #. When component *BB* is in the :ref:`encapsulated set<specC_encapsulation_set>` of component *AA* the applicable interface for variable *B* SHALL be the public interface, and the applicable interface for variable *A* SHALL be the private interface.
 
-    #. It is noted, for the avoidance of doubt, that any other encapsulation relationship between components *AA* and *BB* will not permit connections, meaning that no interface is available between the variables *A* and *B*.
+    It is noted, for the avoidance of doubt, that if components *AA* and *BB* are in each other's :ref:`hidden set<specC_hidden_set>`, there are no applicable interfaces for the variables *A* and *B*.
 
-8.  CellML models MUST only contain :code:`map_variables` elements where the interface of variable *A* and the interface of variable *B* are applicable interfaces.
+8.  CellML models MUST only contain :code:`map_variables` elements where the applicable interfaces for both variables are available.
```
agarny, line 416: Shouldn't this read "... where the applicable interface for both variable is available"? A variable can only have one applicable interface, not several of them...?

src/reference/sectionC_interpretation.inc
MichaelClerx: And we don't need a spec that has super formulaic sentences. As long as it's unambiguous.

src/reference/sectionC_interpretation.inc
MichaelClerx: for _both_ variables, so the applicable interface for A must be available, and the applicable interface for B must be available

src/reference/sectionC_interpretation.inc
MichaelClerx: ok changed it

src/reference/sectionC_interpretation.inc
agarny: Ok, I stand corrected. :)

src/reference/sectionC_interpretation.inc
agarny: Ok, 2-1, so fine with that. :)

agarny:
Ping @kerimoyle.

Issue #261, opened 2020-04-16
3.9.1 Interpretation of encapsulation --> loops


MichaelClerx:
> 3.9. Interpretation of encapsulation elements
> 1. For the purposes of this specification, there SHALL be a “conceptual encapsulation digraph” in which there is EXACTLY one node for every component in the CellML model. **Therefore the encapsulation digraph will not contain any loops.**

I don't understand this "therefore the encapsulation digraph will not contain any loops".

Firstly, we haven't defined the graph yet, so a therefore seems a bit premature.

Secondly, I don't see how it follows from the other rules that there won't be any loops?

MichaelClerx:
OK hold on, it does follow, but it follows from point 1 _and_ 2, not just from 1

MichaelClerx:
This needs fixing actually. 

In B we say that:
- a `comp_ref` can have `comp_ref` children

In C we say that:
- every comp in the model creates a node in the graph
- every child `comp_ref` of a parent `comp_ref` creates an arc in the graph

We're missing the rule saying each `comp_ref` must name a unique component.

PR #268, opened 2020-04-16
Encapsulation loops

MichaelClerx:
Added rule saying the component referenced by a component_ref element must be unique, which prevents cycles in the encapsulation digraph.

See #261

### DIFF
--- a/src/reference/sectionB_elements.inc
+++ b/src/reference/sectionB_elements.inc
@@ -454,6 +454,8 @@ A :code:`component_ref` element information item (referred to in this specificat
 
       The value of the :code:`component` attribute MUST be a valid component reference, as defined in the section :ref:`Component references<specC_component_reference>`.
 
+      The value of the :code:`component` attribute MUST NOT be identical to the value of the :code:`component` attribute on any other :code:`component_ref` element in the :ref:`CellML infoset<specA_cellml_infoset>`.
+
 .. container:: issue-component-ref-child
 
    2. Every :code:`component_ref` element MAY in turn contain one or more :code:`component_ref` element children.
--- a/src/reference/sectionC_interpretation.inc
+++ b/src/reference/sectionC_interpretation.inc
@@ -328,8 +328,7 @@ Interpretation of ``math`` elements
 Interpretation of ``encapsulation`` elements
 --------------------------------------------
 
-1. For the purposes of this specification, there SHALL be a "conceptual encapsulation digraph" in which there is EXACTLY one node for every component in the :ref:`CellML model<specA_cellml_model>`.
-   Therefore the encapsulation digraph will not contain any loops.
+1. For the purposes of this specification, there SHALL be a conceptual "encapsulation digraph" in which there is exactly one node for every component in the :ref:`CellML model<specA_cellml_model>`.
 
 2. Where a :code:`component_ref` element appears as a child of another :code:`component_ref` element, there SHALL be an arc in the encapsulation digraph, and that arc SHALL be from the node corresponding to the component referenced by the parent :code:`component_ref` element, and to the node corresponding to the component referenced by the child :code:`component_ref` element.
 


src/reference/sectionB_elements.inc
```diff
@@ -454,6 +454,8 @@ A :code:`component_ref` element information item (referred to in this specificat
 
       The value of the :code:`component` attribute MUST be a valid component reference, as defined in the section :ref:`Component references<specC_component_reference>`.
```
kerimoyle, line 455: @MichaelClerx are your changes to numbered references in another PR that's waiting, or should we be looking for them here too? Just thinking that this "as defined in" should get a number?

MichaelClerx: They're in https://github.com/cellml/cellml-specification/pull/269

MichaelClerx: So should definitely not be made here or we could get conflicts!

src/reference/sectionB_elements.inc
```diff
@@ -454,6 +454,8 @@ A :code:`component_ref` element information item (referred to in this specificat
...
 
+      The value of the :code:`component` attribute MUST NOT be identical to the value of the :code:`component` attribute on any other :code:`component_ref` element in the :ref:`CellML infoset<specA_cellml_infoset>`.
```
kerimoyle, line 457: I thought that there was a reason for not having statements about value in Chapter 2?  We have all the other regulations about interpreting the value of an attribute in Chapter 3? (eg: names of units, names of connected variables, value of component_1 attributes etc etc)?

MichaelClerx: Ones like this can be checked purely syntactically, and we have them in 2. E.g. the uniqueness of component and variable names

MichaelClerx: Search for "MUST NOT be identical"

MichaelClerx: Statements that go in 3 should be ones that you need an interpretation for. So if you can't (easily) do it with pure syntax rules, but need a bit of interpretation to make sense of the rule (e.g. a digraph or network or whatever), then it goes in 3

kerimoyle: I don't get it ... so why can't we say variable_1 and variable_2 in a mapping can't be equal? or should we be?

MichaelClerx: Some stuff could go in either. For variable_1 and variable_2 I wouldn't have minded if we did some of the stuff syntactically, instead of doing it all through the network definition, but I guess that's a preference?

MichaelClerx: I did think about putting this one in 3 instead. But the similarity to the name uniqueness made me stick it here.
(Also the stuff in 3 starts with "a node for every component in the model", so then the rule would have to be "each node can only correspond to a single component_ref element" or something... so probably better then to rewrite more... This seemed like the minimal effort way to do it!

kerimoyle: Method in the madness but not madness in the method.  As long as there's a reason, I'm ok.

src/reference/sectionC_interpretation.inc
```diff
@@ -333,8 +333,7 @@ Interpretation of ``math`` elements
 Interpretation of ``encapsulation`` elements
 --------------------------------------------
 
-1. For the purposes of this specification, there SHALL be a conceptual "encapsulation digraph" in which there is EXACTLY one node for every component in the :ref:`CellML model<specA_cellml_model>`.
-   Therefore the encapsulation digraph will not contain any loops.
```
kerimoyle, line 337: Do we still say somewhere that it can't have loops? I know it's a consequence of the other points, but "for the avoidance of doubt" yada yada?

kerimoyle: perhaps in (4) below?

MichaelClerx: If you like. You can add it after point 2, cause it follows from 1&2+the rules in chapter 2

But to me it distracted a bit, as the rest of this section is about defining the encapsulated, sibling, and hidden sets

MichaelClerx: Also, if you follow the rules in chapter 2, you have a structure that won't lead to cycles.
So it's already sorted at this point. It's not a rule you need to realise is implied by these points and then check. If you've done chapter 2 right it just can't happen at this point.

MichaelClerx: Could be mentioned in informative spec instead maybe?

hsorby: Having already ruled out loops there seems to be no need to add a reminder statement.  This would also be inconsistent with the rest of the specification.

PR #269, opened 2020-04-16
Changed 'see X' links from `:ref:` to `:numref:`

### LINKED ISSUE #239: Numbers in hyperlinks/references
So that you can still navigate if you print the spec out

MichaelClerx:
Fixes #239.

There's a lot of links that aren't important enough to interrupt the flow of the text with a section number, so have left them as they are.

Also links to tables for some reason can't be numrefs, so haven't touched those.


kerimoyle:
The tables don't have a heading associated with them, it's a manual link, because otherwise they became an entire section each.  I think numref only works for sections ...

Issue #276, opened 2020-04-16
Remove unneeded support for hoverxref


agarny:
Not only we don't make use of it, but it also breaks our build.

MichaelClerx:
Is it not the thing that makes those tooltips appear?

https://pypi.org/project/sphinx-hoverxref/

kerimoyle:
This is what makes the tooltips appear :) And it only breaks our build really recently ... so it's our fault, no theirs!

agarny:
Closing as this should be fixed, not removed. See PR #277.

Issue #240, opened 2020-04-16
Tables invisible or badly formatted in pop-up


MichaelClerx:

kerimoyle:
Note to self: see https://github.com/readthedocs/sphinx-hoverxref/issues/38

Issue #192, opened 2020-04-13
Why is the Latex builder looking for hoverxref?


kerimoyle:
My builds are now showing this error: (seeL https://readthedocs.org/projects/cellml-specification-dev/builds/10828631/)
```
Exception occurred:
  File "/home/docs/checkouts/readthedocs.org/user_builds/cellml-specification-dev/envs/i143_3_10_map_variables_informative/lib/python3.7/site-packages/hoverxref/domains.py", line 50, in _get_docpath
    docpath = builder.get_outfilename(docname)
AttributeError: 'LaTeXBuilder' object has no attribute 'get_outfilename'
```

Since this morning - no changes to config or anything that I know of.  I don't know why the LaTeXBuilder is having anything to do with hoverxref? @hsorby ... any thoughts?

MichaelClerx:
I'm getting this error now too

```
Running Sphinx v3.0.1
making output directory... done

Exception occurred:
  File "/usr/local/lib/python3.7/site-packages/hoverxref/extension.py", line 198, in setup_sphinx_tabs
    for listener_id, function in app.events.listeners.get('html-page-context').items():
AttributeError: 'list' object has no attribute 'items'
The full traceback has been saved in /tmp/sphinx-err-0jktz51h.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [Makefile:45: html] Error 2
```

agarny:
Oh, you are using Sphinx 3.0.1. You want Sphinx 2.4.x (see https://github.com/cellml/cellml-specification/issues/281#issuecomment-614934631).

MichaelClerx:
> Oh, you are using Sphinx 3.0.1. You want Sphinx 2.4.x (see [#281 (comment)](https://github.com/cellml/cellml-specification/issues/281#issuecomment-614934631)).

Thanks!

MichaelClerx:
Does this still happen with the latest hoverref version?

agarny:
Personally, I am now using Sphinx 3.0.1 with sphinx-hoverxref 0.3b1, and it's all working fine.

PR #277, opened 2020-04-16
Remove unneeded support for hoverxref

agarny:
Addresses issue #276.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -28,7 +28,6 @@ Then activate the virtual environment and instal the required packages::
 
   source venv_specification/bin/activate
   pip install sphinx sphinx_rtd_theme
-  pip install -r cellml-specification/requirements.txt
 
 That completes the environment setup required for building the documentation.
 Once the environment is completed, the HTML form of the documentation can be built with the following commands::
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,0 @@
-sphinx-hoverxref
--- a/src/conf.py
+++ b/src/conf.py
@@ -80,7 +80,6 @@ extensions = [
     'sphinx.ext.extlinks',
     'sphinx.ext.autosectionlabel',
     'sphinx.ext.intersphinx',
-    'hoverxref.extension',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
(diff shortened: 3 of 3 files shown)


MichaelClerx:
Is it not the thing that makes those tooltips appear?

https://pypi.org/project/sphinx-hoverxref/

https://sphinx-hoverxref.readthedocs.io/en/latest/

MichaelClerx:
I don't mind disabling it (esp. if it's in beta, as the docs suggest), but I think @kerimoyle likes it?

MichaelClerx:
See #192 and #240 ?

kerimoyle:
Nooooo ... the reason it breaks our build is because of something funky we've done this week ... and it's superuseful for the informative stuff where we don't expect people to have read and digested every word of the normative spec first.

kerimoyle:
Having said that, because our PDF will be built somewhere else, we should turn off online PDF building from the readthedocs site, and simply serve the "real" PDF via our site.  If people want to download the repo and build their own, then that's on them ...

agarny:
Ok, fair enough. Although `src/conf.py` does/did mention "Automagically convert all :ref: blocks to show a tooltip using the hoverxref", I clearly ignored it.

I agree that it's nice to have, so we should fix the problem... somehow. Am therefore going to close this PR and the corresponding issue.

humitos:
Just chiming in here after reading that you had some issues with sphinx-hoverxref extension. Sorry :)

> Nooooo ... the reason it breaks our build is because of something funky we've done this week ...

Probably, most of your issues with the extension should be fixed with the latest release 0.3b1, released ~1h ago.

> and it's superuseful for the informative stuff where we don't expect people to have read and digested every word of the normative spec first.

:heart_eyes: 

> Having said that, because our PDF will be built somewhere else, we should turn off online PDF building from the readthedocs site, and simply serve the "real" PDF via our site.

What's the problem that you are facing with RTD building your PDF? Was it just because of this extension? What would be the difference of building on RTD than in another place?

MichaelClerx:
Thanks!
We just want there to be a single "official" PDF, so we're disabling PDF building on readthedocs. Not a technical issue :-)

PR #284, opened 2020-04-16
Made normative-only the default build.

### LINKED ISSUE #167: Build normative spec. by default?
Should we build the normative spec. by default and the combined informative and normative spec. as an option?

MichaelClerx:
Related to, but does not fully address #167.


MichaelClerx:
This makes everything build without warnings for me, without having to fiddle around with environment variables :D

MichaelClerx:
I think that'd involve changing the makefile, which is not usually recommended for sphinx projects

hsorby:
It's kind of mutating the Sphinx build process even worse that what we are doing already.

Issue #280, opened 2020-04-16
Do we need numbered links for all external references?


kerimoyle:
We have several places where we have embedded the website in our text, and it appears in the reference section too.  Should those external sites have numbered references throughout the text so that the paper reader can find them?

MichaelClerx:
How do you mean, "embedded the website in our text" ?

kerimoyle:
![image](https://user-images.githubusercontent.com/46887220/79511490-80f04800-8093-11ea-8507-2cb83292c6d5.png)

Where we've written it as if it will be read online only, so made a clickable link to the reference itself.  Maybe we should be pointing those links to Chapter 4?

MichaelClerx:
I see your point!

Not 100% sure though, external links like these are clickable from the PDF, at least

kerimoyle:
Yep, but printed versions are suddenly without the most important refs?

MichaelClerx:
Well as you say, they're at the end of the spec. But even that's for completeness mostly, no? Most people won't want to manually type the URL into a browser?

MichaelClerx:
If you can do it without changing the flow too much I see no harm :-)

hsorby:
Can we not add superscripts to the link for the reference section here for PDF builder?

MichaelClerx:
How would you go about that? I'd probably prefer a textual fix (e.g. "see chapter 4")

hsorby:
I don't think we need numbered references I would put it down as a nice to have.

kerimoyle:
Can you see what's there and change as needed?  I don't know how to do any of it with latex for the PDF ...?

PR #287, opened 2020-04-17
I280 numbering references

kerimoyle:
Addresses #280 and #288 

Numbered references throughout the text so that people who are reading on paper can follow them.  

Global numbering is not supported (ie, can't use [1] etc) as that's reserved for local footnotes.  I've used [Ref_1] ... but open to other ideas ...

Rendered version is here: https://cellml-specification-dev.readthedocs.io/en/i280_numbering_references/reference/formal_only/specA2.html#cellml-and-xml

### DIFF
--- a/src/reference/sectionA_definitions.inc
+++ b/src/reference/sectionA_definitions.inc
@@ -7,9 +7,9 @@
 Terminology
 -----------
 
-The keywords "MUST", "MUST NOT", "SHALL", "SHALL NOT", and "MAY" in this document are to be interpreted as described in :rfc2119:`RFC 2119 <>` [Ref_4]_.
+The keywords "MUST", "MUST NOT", "SHALL", "SHALL NOT", and "MAY" in this document are to be interpreted as described in :rfc2119:`RFC 2119 <>` [:ref:`4<ref_rfc2119>`].
 
-The key phrase "information item", as well as any specific type of information item such as an "element information item", are to be interpreted as described in :xml_infoset:`XML Information Set <>` [Ref_6]_.
+The key phrase "information item", as well as any specific type of information item such as an "element information item", are to be interpreted as described in :xml_infoset:`XML Information Set <>` [:ref:`6<ref_xml_infoset>`].
 
 .. _specA_cellml_infoset:
 
@@ -26,7 +26,7 @@ In this specification, the topmost CellML infoset in a hierarchy is referred to 
 .. _specA_namespace:
 
 **Namespace:**
-An XML namespace, as defined in :xml_namespace_1_1:`Namespaces in XML 1.1 <>` [Ref_3]_.
+An XML namespace, as defined in :xml_namespace_1_1:`Namespaces in XML 1.1 <>` [:ref:`3<xml_namespace_1_1>`].
 
 .. _specA_cellml_namespace:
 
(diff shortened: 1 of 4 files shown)


MichaelClerx:
```
src/reference/sectionA_definitions.inc:8: WARNING: Inline interpreted text or phrase reference start-string without end-string.
looking for now-outdated files... none found
pickling environment... done
checking consistency... /home/michael/dev/cellml/specification/src/reference/formal_sectionD.rst:5: WARNING: Citation [Ref_2] is not referenced.
/home/michael/dev/cellml/specification/src/reference/formal_sectionD.rst:9: WARNING: Citation [Ref_4] is not referenced.
```

kerimoyle:
Thanks!

MichaelClerx:
> Thanks!

I've got a fix for some of those if you like :D

hsorby:
Would named references be better here than using numbers?

kerimoyle:
Yeah, I wondered about that.  They were originally named (using the same terms as their references in the conf.py file).  The trouble is that the names are what are displayed, and they were all quite long with no obvious way to shorten them?

MichaelClerx:
Or we could just manually add a "``(see :numref:`{number}. {name}<chapter 4>`)``" everywhere?

hsorby:
> The trouble is that the names are what are displayed,

I thought so I don't want to see that particularly.

MichaelClerx:
> Or we could just manually add a "`` (see :numref:`{number}. {name}<chapter 4>`) ``" everywhere?

@kerimoyle would this be ok? I don't find the "[Ref 7]" links everywhere very pretty or intuitive

kerimoyle:
What would your {name} be? The title of the reference? Won't that be really long?

MichaelClerx:
Chapter 4. Just "see 4. References"

kerimoyle:
Hmm.  Let me try some stuff.

MichaelClerx:
I'm beginning to think this issue 

- users who print the spec don't get a numbered link to external hyperlinks they'll never manually type into a browser anyway

isn't a big enough issue to warrant the kind of changes it seems to imply?

MichaelClerx:
If it interferes with readability for PDF or HTML users I don't really want to do it

kerimoyle:
Yeah, I get it, I just have an idea I want to try.  If that doesn't work out will ditch it.

MichaelClerx:
Thanks!

MichaelClerx:
I don't mind how that looks!

kerimoyle:
Gosh, high praise indeed :D

MichaelClerx:
:D 

It gets worse! Should we renumber so the refs are in order?

kerimoyle:
They're currently alphabetical by title?

MichaelClerx:
ok!

kerimoyle:
See, now you're getting delirious ... :)

agarny:
It started a bit earlier. He's sleep deprived.

src/reference/sectionA_definitions.inc
```diff
@@ -50,7 +50,7 @@ Any information item in the :ref:`CellML namespace<specA_cellml_namespace>`.
 .. _specA_basic_latin_alphabetical_character:
 
 **Basic Latin alphabetical character:**
-A Unicode character in the range :unicode:`0041` to :unicode:`005A` or in the range :unicode:`0061` to :unicode:`007A`.
+A Unicode [:ref:`5<ref_unicode>`] character in the range :unicode:`0041` to :unicode:`005A` or in the range :unicode:`0061` to :unicode:`007A`.
```
hsorby, line 53: I think the `Unicode` word here should be a link.

MichaelClerx:
I'd also prefer order of appearance, but I'll echo Alan's "meh"

> meh!

src/reference/sectionA_definitions.inc
kerimoyle: Good idea.

kerimoyle: Done in 91f99aa

MichaelClerx:
Click the button! Click the button! Quick!

kerimoyle:
Just let me check that the mehs have been fixed ...

MichaelClerx:
> Just let me check that the mehs have been fixed ...

Beautiful! I rescind my echoed meh!

Issue #154, opened 2020-04-02
List of required latex packages?


MichaelClerx:
I've got a lot on my system already, so can't give a full list, but had to install

- latexmk
- fncychap
- tabulary 
- tocloft

to get the pdf docs to build

hsorby:
These are all the packages that I had to install from a minimal installation of ``BasicTex`` on macOS (using the ``tlmgr`` application):

- latexmk
- fncychap
- titlesec
- tabulary
- varwidth
- framed
- wrapfig
- capt-of
- needspace
- tocloft
- textgreek
- babel-greek
- greek-fontenc
- cbfonts
- courier
- helvetic
- fontspec

Not sure if that last one helped me out. I also had to install some fonts for the ``xelatex`` that we use for UTF-8 support

- FreeSans
- FreeSerif
- FreeMono

MichaelClerx:
Those fonts are in
- texlive-gnu-freefont
if you're using texlive

MichaelClerx:
Should we incorporate this info into README.md ?

agarny:
I guess it wouldn't harm. (I remember also having problems at the beginning and ended up installing everything from texlive.)

PR #294, opened 2020-04-17
Updated requirements and gitignore file.

### LINKED ISSUE #293: Requirements file
- Seems we need at least sphinx 2 now to compile the PDF, but not sphinx 3, as that creates an issue with `sphinx-hoverxref`. So... `sphinx>2<3` ?
- I needed a `pip install sphinx_rtd_theme` as well

MichaelClerx:
Closes #293.

With these (and #154) it works for me in a virtual environment

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .vscode/
 .DS_Store
 build/
+venv/
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .DS_Store
 build/
 venv/
+venv_specification/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
 .vscode/
 .DS_Store
 build/
+venv/
+venv_specification/
```
agarny, line 5: Not sure I would add those. They are very specific to one's set up. Personally, I have all of my virtual environments under `~/virtualenv`, so not within a repo as such.

MichaelClerx: We literally tell people to use `venv_specification` in the README, 

Less of a case for `venv`, except that it's pretty common

PR #315, opened 2020-04-20
I314 numref hoverxref

### LINKED ISSUE #314: It would be nice to have hoverxref working on all internal links...
At the moment, anything with a `numref` link is not recognised by hoverxref.

kerimoyle:
Closes #314 

... but ... this currently is proof of concept as it requires a branch from the hoverxref repo.  It can be merged now if we're desperate, or we can wait for it to be included in their update.

See change in the requirements file that would need to be reverted.  

Rendered docs are here: https://cellml-specification-dev.readthedocs.io/en/i314_numref_hoverxref/reference/formal_and_informative/specB09.html

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 sphinx>2,<3
-sphinx-hoverxref
+git+git://github.com/readthedocs/sphinx-hoverxref.git
+# sphinx-hoverxref
 sphinx_rtd_theme
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 sphinx>2,<3
-git+git://github.com/readthedocs/sphinx-hoverxref@releases/tag/0.4b1#egg=sphinx-hoverxref
+git+git://github.com/readthedocs/sphinx-hoverxref@releases/tag/0.4b1
 sphinx_rtd_theme
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 sphinx>2,<3
-git+git://github.com/readthedocs/sphinx-hoverxref@releases/tag/0.4b1
+git+git://github.com/readthedocs/sphinx-hoverxref@releases/tag/v0.4b1#egg=sphinx-hoverxref
 sphinx_rtd_theme
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 sphinx>2,<3
-git+git://github.com/readthedocs/sphinx-hoverxref.git@master
+# git+git://github.com/readthedocs/sphinx-hoverxref.git@master
+sphinx-hoverxref
 sphinx_rtd_theme
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 sphinx>2,<3
-git+git://github.com/readthedocs/sphinx-hoverxref@releases/tag/v0.4b1#egg=sphinx-hoverxref
+git+git://github.com/readthedocs/sphinx-hoverxref@v0.4b1#egg=sphinx-hoverxref
 sphinx_rtd_theme
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 sphinx>2,<3
-# git+git://github.com/readthedocs/sphinx-hoverxref.git@master
 sphinx-hoverxref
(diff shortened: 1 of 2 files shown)


kerimoyle:
Hey @hsorby and @agarny ... am I ok to merge this?  The fix has been made in the master repo of the hoverxref site now.

requirements.txt
```diff
@@ -1,3 +1,3 @@
 sphinx>2,<3
-sphinx-hoverxref
+git+git://github.com/readthedocs/sphinx-hoverxref.git@master
```
agarny, line 2: Hmm... not keen on this. `master` seems to be a wip type of branch. So, who knows, it may at some point contain something that could break things for us. You say that they have fixed the issue numref. Is that this commit: https://github.com/readthedocs/sphinx-hoverxref/commit/3558a67e56bf27f08d52425bbcd1d63989c6224e? If so, and if anything, we might want to point to that one rather than `master` itself. However, my preference would on using an official version of sphinx-hoverxref.

kerimoyle: Yes, that's the commit.  I've tried using just the plain `sphinx-hoverxref` requirement in the file, and it didn't seem to be building it in properly (either that or the cache clearing / readthedocs wiping wasn't being very thorough?).  I'm not sure what the "official" sphinx-hoverxref would point to other than the master branch though?

kerimoyle: I guess I could ask when the next release will be (seems to be quite frequently) so then we can tie it to a tag - that's probably cleaner?

agarny: The official sphinx-hoverxref version is `0.3b1` (as confirmed by my local copy of sphinx-hoverxref and https://github.com/readthedocs/sphinx-hoverxref/releases). So, yes, ideally we would ask them for a new release with the numref fix and keep using `sphinx-hoverxref` rather than `git+git://github.com/readthedocs/sphinx-hoverxref.git@master` or something else.

kerimoyle: They're aiming for another release next week ... fingers crossed :)

agarny: Great. Ping us again in a week's time. :)

kerimoyle:
Discussion here re getting release version with our changes: https://github.com/readthedocs/sphinx-hoverxref/pull/71

kerimoyle:
New release all done :) @agarny all good now?

requirements.txt
```diff
@@ -1,3 +1,3 @@
 sphinx>2,<3
-sphinx-hoverxref
+git+git://github.com/readthedocs/sphinx-hoverxref@0.4b1#egg=sphinx-hoverxref
```
agarny, line 2: Why use the GitHub version since `sphinx-hoverxref` has been updated to 0.4b1 on PyPI? IOW, the original requirement should be good enough.

kerimoyle: See your earlier point https://github.com/cellml/cellml-specification/pull/315#discussion_r416293885 which is why this PR has been waiting in the first place ... ?

agarny: ?? Back then, the official version of `sphinx-hoverxref` was `0.3b1`, which wasn't working for us, hence indeed my suggestion to use a specific revision of `sphinx-hoverxref` from their GitHub repo (rather than the version on the `master` branch, which might be broken or something while I would expect a release to be fine; as per my [comment](https://github.com/cellml/cellml-specification/pull/315#discussion_r416293885)). Either way, now, the official version of `sphinx-hoverxref` is `0.4b1`, which is what we want. So, I would prefer to use the official version (as per my [original](https://github.com/cellml/cellml-specification/pull/315#discussion_r416293885) and [second](https://github.com/cellml/cellml-specification/pull/315#discussion_r416301541) comments).

kerimoyle: Ah, ok, I get it.  I thought you wanted it pinned to a release.  I'll change it now.

kerimoyle:
Right.... done ... @agarny ?

kerimoyle:
There's no CI on this spec repo??  And it's working fine, see rendered version here: https://cellml-specification-dev.readthedocs.io/en/i314_numref_hoverxref/reference/formal_and_informative/specC01_interpretation_of_imports.html (hover over the numbered links)

agarny:
Oops, my bad, sorry! I was in libCellML mode... 🥴

Issue #331, opened 2020-05-26
Update error enum bookmarks to reflect libCellML


kerimoyle:
The highlighting for users of libCellML is based on enum-named tags next to the relevant sections.  These need to be coordinated with those used in libCellML so that the urls from there will link to the correct place here.  See https://github.com/cellml/libcellml/pull/604

kerimoyle:
Closed by #332

PR #332, opened 2020-05-26
I331 update error tags for highlighting

kerimoyle:
Addresses #331 
Because there is more than one error code in libCellML that points to one section, we have to use numbers not enum codes.  It makes it more flexible for use with other users of CellML too.


kerimoyle:
Ready for review please - should be super quick! @agarny @hsorby @nickerso

